PR #6, opened 2020-12-28
Stream practice by ok

Ketch92:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,17 @@
...
 
-public class CandidateValidator {
+import java.util.function.Predicate;
+import model.Candidate;
 
+public class CandidateValidator implements Predicate<Candidate> {
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] yearsInUkraine = candidate.getPeriodsInUkr().split("-");
+        int periodInUkraine = Integer.parseInt(yearsInUkraine[1])
```
RomanDubovskyi, line 10: let's make "Ukrainia", 10 and 35 constants

src/main/java/practice/StreamPractice.java
```diff
@@ -14,18 +40,28 @@
      * "Can't get min value from list: method_input_list"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(s -> Arrays.stream(s.split(",")))
+                .mapToInt(Integer::parseInt)
+                .filter(i -> i % 2 == 0)
+                .min()
+                .orElseThrow(() -> new RuntimeException("Can't get min value from list"));
```
RomanDubovskyi, line 48: let's add list to exception message

src/main/java/practice/StreamPractice.java
```diff
@@ -14,18 +40,28 @@
...
      * Then return the average of all odd numbers or throw NoSuchElementException.
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(i -> i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i))
+                .filter(i -> i % 2 != 0)
+                .mapToDouble(i -> i)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
RomanDubovskyi, line 62:
```suggestion
                .average()
                .get()
```
get will thro NoSuchElement exception by default

src/main/java/practice/StreamPractice.java
```diff
@@ -50,32 +90,27 @@ public Double getOddNumsAverage(List<Integer> numbers) {
...
-     * "2002-2015" For now we don't care if that was last 10 or not. We want to reuse our
-     * validation in future, so
-     * let's write our own impl of Predicate parametrized with Candidate in CandidateValidator.
-     */
-    public static List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex().equals(People.Sex.WOMEN)
+                        && p.getAge() >= femaleAge
+                        && !p.getCats().isEmpty())
```
RomanDubovskyi, line 110:
```suggestion
        return peopleList.stream()
                .filter(p -> p.getSex().equals(People.Sex.WOMEN)
                        && p.getAge() >= femaleAge)
```
+ try to use only single map and single flatMap operation in this task

RomanDubovskyi (changes requested):

src/main/java/practice/CandidateValidator.java
Ketch92: Ok, also moved 1 and 0 to constants.

src/main/java/practice/StreamPractice.java
Ketch92: Sure

src/main/java/practice/StreamPractice.java
Ketch92: humm.... I just thought that we want to define it explicitly, sure why not)

src/main/java/practice/StreamPractice.java
Ketch92: Nice, thx)

src/main/java/practice/StreamPractice.java
Ketch92: Ok, I just run the tests and it fails one of them if I want to change the message. I reversed the change and left it as it is. 
Or don't I got something right?

RomanDubovskyi: this should work 

src/main/java/practice/StreamPractice.java
RomanDubovskyi: that's predictable behavior that's why there is no need to explicitly declare it

src/main/java/practice/StreamPractice.java
```diff
@@ -14,18 +38,28 @@
      * "Can't get min value from list: method_input_list"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(s -> Arrays.stream(s.split(",")))
+                .mapToInt(Integer::parseInt)
+                .filter(i -> i % 2 == 0)
+                .min()
+                .orElseThrow(() -> new RuntimeException("Can't get min value from list"));
```
RomanDubovskyi, line 46: left comment how to fix

src/main/java/practice/StreamPractice.java
```diff
@@ -50,32 +88,26 @@ public Double getOddNumsAverage(List<Integer> numbers) {
...
-     * "2002-2015" For now we don't care if that was last 10 or not. We want to reuse our
-     * validation in future, so
-     * let's write our own impl of Predicate parametrized with Candidate in CandidateValidator.
-     */
-    public static List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex().equals(People.Sex.WOMEN)
+                        && p.getAge() >= femaleAge
+                        && !p.getCats().isEmpty())
```
RomanDubovskyi, line 108:
```suggestion
```
we don't need the check,  flatMap gonna handle this situation for us)

RomanDubovskyi (changes requested):
nice, left 2 minor comments, rest is good)

src/main/java/practice/StreamPractice.java
Ketch92: Ok, thx. I found my mistake

src/main/java/practice/StreamPractice.java
Ketch92: Thanks)

src/main/java/practice/StreamPractice.java
Ketch92: Cool)

RomanDubovskyi (approved):

PR #17, opened 2020-12-29
Hw streams

Likh-Alex:


src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +52,11 @@ public int findMinEvenNumber(List<String> numbers) {
      * Then return the average of all odd numbers or throw NoSuchElementException.
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(number -> number % 2 == 0 ? numbers.get(number) : numbers.get(number) - 1)
+                .filter(number -> number % 2 != 0)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
Ketch92, line 59: you can you only get() it throws needed exception by default

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +86,14 @@ public Double getOddNumsAverage(List<Integer> numbers) {
...
     public List<People> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<People> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getAge() >= fromAge
+                        && (person.getSex().equals(People.Sex.MAN)
+                        && person.getAge() <= maleToAge)
+                        || person.getAge() >= fromAge
+                        && (person.getSex().equals(People.Sex.WOMEN)
+                        && person.getAge() <= femaleToAge))
```
Ketch92, line 95: you can use the ternary operator here

src/main/java/practice/StreamPractice.java
```diff
@@ -59,23 +102,12 @@ public Double getOddNumsAverage(List<Integer> numbers) {
...
-     * let's write our own impl of Predicate parametrized with Candidate in CandidateValidator.
-     */
-    public static List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getSex().equals(People.Sex.WOMEN)
+                        && person.getAge() >= femaleAge)
+                .map(people -> people.getCats())
+                .flatMap(cats -> cats.stream())
+                .map(cat -> cat.getName())
```
Ketch92, line 110: you can use only one map operator

src/main/java/practice/StreamPractice.java
```diff
@@ -1,11 +1,35 @@
...
+     * and live in urk for 10 years. For the last requirement use field periodsInUkr,
+     * which has following view:
+     * "2002-2015" For now we don't care if that was last 10 or not. We want to reuse our
+     * validation in future, so
+     * let's write our own impl of Predicate parametrized with Candidate in CandidateValidator.
+     */
+    public static List<String> validateCandidates(List<Candidate> candidates) {
+        return candidates.stream()
+                .filter(new CandidateValidator())
+                .map(candidate -> candidate.getName())
```
Ketch92, line 28: can you change the method reference?

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +38,12 @@
      * "Can't get min value from list: method_input_list"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(stringNum -> Arrays.stream(stringNum.split(",")))
+                .mapToInt(s -> Integer.parseInt(s))
```
Ketch92, line 43: also method reference can be applied

Ketch92 (changes requested):

src/main/java/practice/StreamPractice.java
Likh-Alex: Done, thank you for an idea)

src/main/java/practice/StreamPractice.java
Likh-Alex: Done, refactored to use one map and one flatMap

src/main/java/practice/StreamPractice.java
Likh-Alex: I think in this particular case ternary operator shortens the code but complicates the readability. Anyway - done, it is good to practice)

src/main/java/practice/StreamPractice.java
Likh-Alex: Done, I just liked to use lambdas as they are clearer to me)

src/main/java/practice/StreamPractice.java
Likh-Alex: Done, as well)

Ketch92 (approved):

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +38,12 @@
      * "Can't get min value from list: method_input_list"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(stringNum -> Arrays.stream(stringNum.split(",")))
+                .mapToInt(Integer::parseInt)
+                .filter(num -> num % 2 == 0)
+                .min()
+                .orElseThrow(() -> new RuntimeException("Can't get min value from list"));
```
RomanDubovskyi, line 46:
```suggestion
                .orElseThrow(() -> new RuntimeException("Can't get min value from list: " + numbers));
```
let's make our message more informative

RomanDubovskyi (approved):
fix message, rest seems to be good👍

src/main/java/practice/StreamPractice.java
Likh-Alex: Done

PR #41, opened 2021-01-07
zero commit

entertainment-veks:


RomanDubovskyi:
try to use more informative commit and PR names
image

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,24 @@
 package practice;
 
-public class CandidateValidator {
+import java.util.function.Predicate;
+import model.Candidate;
 
+public class CandidateValidator implements Predicate<Candidate> {
+    @Override
+    public boolean test(Candidate candidate) {
```
RomanDubovskyi, line 8: this check may be simplified and it's better to declare nationality and age as a constant to avoid magic numbers/string

src/main/java/practice/StreamPractice.java
```diff
@@ -1,20 +1,39 @@
...
      * Given list of strings where each element contains 1+ numbers:
      * input = {"5,30,100", "0, 22, 7", ...}
      * return min integer value. One more thing - we're interested in even numbers.
      * If there is no needed data throw RuntimeException with message
      * "Can't get min value from list: method_input_list"
      */
+
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        OptionalInt result =
```
RomanDubovskyi, line 24: you can use method orElseThrow in stream and you won't need to declare a separate variable

src/main/java/practice/StreamPractice.java
```diff
@@ -1,20 +1,39 @@
...
+                numbers.stream()
+                .mapToInt(
```
RomanDubovskyi, line 26: you don't need to call min twice in stream

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +42,19 @@ public int findMinEvenNumber(List<String> numbers) {
      * Then return the average of all odd numbers or throw NoSuchElementException.
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        indexOfGetOddNumsAverage = 0;
+        OptionalDouble result = numbers.stream()
```
RomanDubovskyi, line 46: same here, use getAsDouble instead of creating a new optional variable and checking it

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +84,23 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      */
     public List<People> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<People> peopleList) {
-        return Collections.emptyList();
+
+        Predicate<People> getWorkablePeoplePredicator = people -> {
```
RomanDubovskyi, line 88: it will be shorter if you write this check without if and inside of the filter method, you can use just the logical operators

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +109,12 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<People> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(people -> people.getSex().equals(People.Sex.WOMEN))
+                .filter(people -> people.getAge() >= femaleAge)
```
RomanDubovskyi, line 114: checklist

RomanDubovskyi (approved):

PR #44, opened 2021-04-20
added solution

freedmanMain:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,78 @@
...
-     */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream
+                .range(0, numbers.size())
+                .map(i -> i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i))
+                .filter(i -> i % 2 != 0)
+                .mapToDouble(d -> d)
+                .average()
+                .getAsDouble();
```
mankokolya, line 31: You forget to throw Exception here

mankokolya (changes requested):

src/main/java/practice/StreamPractice.java
freedmanMain: If a value is present in this OptionalDouble, returns the value, otherwise throws NoSuchElementException.

mankokolya (approved):

Resci (approved):

Resci (approved):

PR #73, opened 2021-04-27
First Solution

Ihor-Nykyforov:


KseniiaMakarova (changes requested):
test

Anatolii-Zarytskyi (approved):
Good job)

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,21 @@
...
 
-public class CandidateValidator {
+import java.util.function.Predicate;
+import model.Candidate;
 
+public class CandidateValidator<T extends Candidate> implements Predicate<T> {
+    private static final String DELIMITER = "-";
+    private static final String NATIONALITY = "Ukrainian";
+    private static final int REQUIRED_AGE = 35;
+    private static final int REQUIRED_YEARS_OF_LIVING_IN_COUNTRY = 10;
```
Resci, line 10:
```suggestion
    private static final int MIN_REQUIRED_AGE = 35;
    private static final int MIN_REQUIRED_YEARS_OF_LIVING_IN_COUNTRY = 10;
```

src/main/java/practice/StreamPractice.java
```diff
@@ -1,20 +1,32 @@
...
      */
+
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(s -> Arrays.stream(s.split(DELIMITER)))
+                .mapToInt(Integer::parseInt)
+                .filter(integer -> integer % 2 == 0)
+                .min()
+                .orElseThrow(() -> new RuntimeException(MESSAGE));
```
Resci, line 29: You use message only in 1 place, let's remove const

Resci (approved):

PR #78, opened 2021-04-27
Implemented methods according to conditions in README.

Kharlanava:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,71 @@
...
-     * for example, `Arrays.asList( new People(«Victor», 16, Sex.MAN),
-     * new People(«Helen», 42, Sex.WOMEN))`,
-     * select from the List only men whose age is from `fromAge` to `toAge` inclusively.
-     * <p>
-     * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
-     */
     public List<People> selectMenByAge(List<People> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(human -> human.getSex().equals(People.Sex.MAN)
```
veronika1shcherbyna, line 34: From checklist: _If you have a long boolean expression you may create your own implementation of Predicate and use it in your filter._ 
So you can create your own Predicate object :)

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,71 @@
...
-     * or to `femaleToAge` (for women) inclusively.
-     * <p>
-     * Example: select people of working age
-     * (from 18 y.o. and to 60 y.o. for men and to 55 y.o. for women inclusively).
-     */
     public List<People> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<People> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(human -> (human.getSex().equals(People.Sex.MAN)
```
veronika1shcherbyna, line 43: the same here

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,71 @@
...
 
-    /**
-     * Given a List of `People` instances (having `name`, `age`, `sex` and `List of cats` fields,
-     * and each `Cat` having a `name` and `age`),
-     * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
-     */
     public List<String> getCatsNames(List<People> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        List<People> validPeople = peopleList.stream()
+                .filter(human -> human.getSex().equals(People.Sex.WOMEN)
```
veronika1shcherbyna, line 54: and you can use Predicate here

veronika1shcherbyna (changes requested):

veronika1shcherbyna (approved):

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,90 @@
...
+            }
+        };
+        List<People> validPeople = peopleList.stream()
+                .filter(selectPeople)
+                .collect(Collectors.toList());
+        return validPeople.stream()
+                .map(People::getCats)
+                .flatMap(Collection::stream)
+                .map(Cat::getName)
+                .collect(Collectors.toList());
```
Resci, line 80: We can merge it into one stream with 1 map operation and 1 flatMap operation. Let's think about how we can do that

Resci (changes requested):

src/main/java/practice/StreamPractice.java
Kharlanava: Done.

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,88 @@
...
+            public boolean test(People human) {
+                return human.getSex().equals(People.Sex.WOMEN)
+                        && human.getAge() >= femaleAge;
+            }
+        };
+        return peopleList.stream()
+                .filter(selectPeople)
+                .map(People::getCats)
+                .flatMap(Collection::stream)
+                .map(Cat::getName)
```
Resci, line 77:
```suggestion
                .filter(selectPeople)
                .flatMap(people -> people.getCats().stream())
                .map(Cat::getName)
```

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,88 @@
...
-     * The requirements are: person should be older than 35 y, should be allowed to vote,
-     * have nationality - 'Ukrainian'
-     * and live in urk for 10 years. For the last requirement use field periodsInUkr,
-     * which has following view: "2002-2015"
-     * We want to reuse our validation in future, so let's write our own impl of Predicate
-     * parametrized with Candidate in CandidateValidator.
-     */
-    public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+    public static List<String> validateCandidates(List<Candidate> candidates) {
```
Resci, line 81: Why static?

Resci (changes requested):

src/main/java/practice/StreamPractice.java
Kharlanava: Because unit tests call this method from the class. So we can not use non-static method in static context.

Resci (approved):

PR #82, opened 2021-04-27
HW Streams

veronika1shcherbyna:


src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +34,11 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(index -> index % 2 != 0 ? numbers.get(index) - 1 : numbers.get(index))
+                .filter(number -> number % 2 != 0)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
Kharlanava, line 41: You may use also getAsDouble() here. It throws NoSuchElementException if there is no appropriate value

Kharlanava (approved):

Resci (approved):

PR #83, opened 2021-04-27
Done Homework.

DeStinyShell:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,17 @@
...
+    private static final int REQUIRED_AGE = 35;
+    private static final String REQUIRED_NATIONALITY = "Ukrainian";
+
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= REQUIRED_AGE && candidate.isAllowedToVote()
+                && candidate.getNationality().equals(REQUIRED_NATIONALITY)
+                && Math.abs(Integer.parseInt(candidate.getPeriodsInUkr().split("-")[0])
+                - Integer.parseInt(candidate.getPeriodsInUkr().split("-")[1])) >= 10;
+    }
```
LayRipper, line 16: You can avoid Math.abs() method by interchanging the subtractor
10 should be denoted as constant as well

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +37,12 @@
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(string -> string.split(","))
+                .flatMapToInt(array -> Arrays.stream(array).mapToInt(Integer::parseInt))
```
LayRipper, line 42: 2 map operations are sufficient here

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +51,12 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .mapToObj(index -> index % 2 != 0 ? numbers.get(index) - 1 : numbers.get(index))
+                .filter(index -> index % 2 != 0)
+                .mapToDouble(Double::valueOf)
```
LayRipper, line 57: Let's use 1 map operation instead of two, and we should be careful with Double values comparison as != may cause bugs so let's use a safer way to compare.
```suggestion
                .mapToDouble(index -> index % 2 != 0 ? numbers.get(index) - 1 : numbers.get(index))
                .filter(index -> index % 2 != 0)
```

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +68,11 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<People> selectMenByAge(List<People> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getSex().equals(People.Sex.MAN)
+                        && person.getAge() >= fromAge
+                        && person.getAge() <= toAge)
```
LayRipper, line 74: we should create an anonymous predicate class here as per checklist

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +87,13 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      */
     public List<People> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<People> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> {
+                    if (person.getSex().equals(People.Sex.MAN)) {
+                        return person.getAge() >= fromAge && person.getAge() <= maleToAge;
+                    }
+                    return person.getAge() >= fromAge && person.getAge() <= femaleToAge;
```
LayRipper, line 95: here as well

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +68,11 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<People> selectMenByAge(List<People> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getSex().equals(People.Sex.MAN)
```
LayRipper, line 72: It is generally better to compare Enums via == rather than equals()

src/main/java/practice/StreamPractice.java
```diff
@@ -59,22 +102,11 @@ public Double getOddNumsAverage(List<Integer> numbers) {
...
-     * We want to reuse our validation in future, so let's write our own impl of Predicate
-     * parametrized with Candidate in CandidateValidator.
-     */
-    public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getSex().equals(People.Sex.WOMEN)
+                        && person.getAge() >= femaleAge)
+                .map(People::getCats)
+                .flatMap(cats -> cats.stream().map(Cat::getName))
```
LayRipper, line 109:
```suggestion
                .flatMap(cats -> cats.getCats().stream())
                .map(Cat::getName)
```

LayRipper (changes requested):

LayRipper (approved):

src/main/java/practice/StreamPractice.java
```diff
@@ -1,11 +1,37 @@
...
+     * have nationality - 'Ukrainian'
+     * and live in urk for 10 years. For the last requirement use field periodsInUkr,
+     * which has following view: "2002-2015"
+     * We want to reuse our validation in future, so let's write our own impl of Predicate
+     * parametrized with Candidate in CandidateValidator.
+     */
+
+    public static List<String> validateCandidates(List<Candidate> candidates) {
+        return candidates.stream()
+                .filter(new practice.CandidateValidator())
```
Resci, line 29: Why practice?

Resci (changes requested):

LayRipper (approved):

src/main/java/practice/StreamPractice.java
LayRipper: ой, sorry

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +40,12 @@
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMapToInt(string -> Arrays.stream(string.split(","))
+                        .mapToInt(Integer::parseInt))
```
Resci, line 45:
```suggestion
                .flatMapToInt(string -> Arrays.stream(string.split(","))
                .mapToInt(Integer::parseInt))
```

Resci (approved):

PR #93, opened 2021-04-27
Streams. Ready for review.

MakeDreamCode:


src/main/java/practice/StreamPractice.java
```diff
@@ -29,9 +22,10 @@
...
         OptionalInt min = numbers.stream()
-                                 .flatMap(stringNumber -> Arrays.stream(stringNumber.split(DELIMITER)))
+                                 .flatMap(stringNumber ->
+                                         Arrays.stream(stringNumber.split(DELIMITER)))
                                  .mapToInt(Integer::parseInt)
-                                 .filter(n -> n % 2 == 0)
+                                 .filter(number -> number % 2 == 0)
                                  .min();
         if (min.isEmpty()) {
             throw new RuntimeException(ERROR_MESSAGE + numbers.toString());
```
mmelniichuk: It's seemed toString() is not necessary method.
And you can call orElseThrow() right after method min()...
but it is optional, your solution is right

src/main/java/practice/StreamPractice.java
```diff
@@ -62,7 +57,8 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      */
     public List<People> selectMenByAge(List<People> peopleList, int fromAge, int toAge) {
         return peopleList.stream()
-                         .filter(p -> p.getSex() == People.Sex.MAN && p.getAge() >= fromAge && p.getAge() <= toAge)
+                         .filter(people -> people.getSex() == People.Sex.MAN
+                                 && people.getAge() >= fromAge && people.getAge() <= toAge)
```
mmelniichuk, line 61: If you have a long boolean expression you may create your own implementation of Predicate and use it in your filter. - checklist

src/main/java/practice/StreamPractice.java
```diff
@@ -78,24 +74,12 @@ public Double getOddNumsAverage(List<Integer> numbers) {
...
-
         return peopleList.stream()
-                         .filter(p -> p.getAge() >= fromAge)
-                         .filter(p -> (p.getSex() == People.Sex.WOMEN && p.getAge() <= femaleToAge)
-                                 || (p.getSex() == People.Sex.MAN && p.getAge() <= maleToAge))
+                         .filter(people -> people.getAge() >= fromAge)
+                         .filter(people -> (people.getSex() == People.Sex.WOMEN
+                                 && people.getAge() <= femaleToAge)
+                                 || (people.getSex() == People.Sex.MAN
+                                 && people.getAge() <= maleToAge))
```
mmelniichuk, line 82: Your commented solution (deleted) about your own implementation of Predicate was right. According checkstyle it would be better to implement Predicate (example Predicate<People> isWorkablePeople = pass your filter-expression here). and one filter() method would be enough

mmelniichuk (changes requested):
Good job! But I have some recommendations to change

src/main/java/practice/StreamPractice.java
MakeDreamCode: Done.

src/main/java/practice/StreamPractice.java
MakeDreamCode: Done.

src/main/java/practice/StreamPractice.java
MakeDreamCode: replaced by .orElseThrow(()
                        -> new RuntimeException("Can't get min value from list: "
                        + numbers));

mmelniichuk (approved):

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +54,16 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<People> selectMenByAge(List<People> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        Predicate<People> mansForArmy = new Predicate<People>() {
+            @Override
+            public boolean test(People people) {
+                return people.getSex() == People.Sex.MAN
+                        && people.getAge() >= fromAge && people.getAge() <= toAge;
+            }
```
KseniiaMakarova, line 62: let's write it as lambda

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +54,16 @@ public Double getOddNumsAverage(List<Integer> numbers) {
...
+        };
```
KseniiaMakarova, line 63:
```suggestion
        Predicate<People> mansForArmy = person -> person.getSex() == People.Sex.MAN
                        && person.getAge() >= fromAge && person.getAge() <= toAge;
```

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +78,19 @@ public Double getOddNumsAverage(List<Integer> numbers) {
...
+        Predicate<People> isWorkablePeople = new Predicate<People>() {
+            @Override
+            public boolean test(People people) {
+                return people.getAge() >= fromAge
+                        && ((people.getSex() == People.Sex.WOMEN
+                        && people.getAge() <= femaleToAge)
+                        || (people.getSex() == People.Sex.MAN
+                        && people.getAge() <= maleToAge));
+            }
+        };
```
KseniiaMakarova, line 90: same here

KseniiaMakarova (approved):

PR #95, opened 2021-05-24
added solution to steam practice

mtv1301:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +28,15 @@
...
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        try {
+            return Stream.of(numbers)
+                    .flatMap(Collection::stream)
+                    .flatMap(str -> Arrays.stream(str.split(",")))
+                    .filter(num -> Integer.parseInt(num) % 2 == 0)
+                    .mapToInt(Integer::parseInt).sorted().findFirst().getAsInt();
+        } catch (RuntimeException e) {
```
boroda4436, line 37: please try to use `orElseThrow`

src/main/java/practice/StreamPractice.java
```diff
@@ -23,9 +45,12 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        OptionalDouble average = IntStream.range(0, numbers.size())
+                .map(i -> i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i))
+                .peek(System.out::println)
+                .filter(num -> num % 2 != 0).average();
+        return average.getAsDouble();
```
boroda4436, line 52: Feel free to return the result directly without instantiating the new variable.

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +110,17 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        candidates.forEach(candidate -> {
+            String[] split = candidate.getPeriodsInUkr().split("-");
+            candidate.setPeriodsInUkr(String.valueOf(Integer.parseInt(split[1]) - Integer.parseInt(split[0])));
+        });
+        return candidates.stream()
```
boroda4436, line 117: Here you should use your class CandidateValidator

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +110,17 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        candidates.forEach(candidate -> {
```
boroda4436, line 113: It should be enough to use only one stream and do not use this foreach here

boroda4436 (changes requested):

PR #102, opened 2021-06-28
Hw stream by skvortsov

Dhakmar:


aaananas (approved):

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,74 @@
...
 public class StreamPractice {
-    /**
-     * Given list of strings where each element contains 1+ numbers:
-     * input = {"5,30,100", "0,22,7", ...}
-     * return min integer value. One more thing - we're interested in even numbers.
-     * If there is no needed data throw RuntimeException with message
-     * "Can't get min value from list: < Here is our input 'numbers' >"
-     */
+    private static final String NUMBER_SEPARATOR = ",";
+    private static final int DECREMENT = 1;
```
Sofasmile, line 15: it's optional to create constant for 1

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,74 @@
...
+    private static final CandidateValidator CANDIDATE_VALIDATOR = new CandidateValidator();
```
Sofasmile, line 16: let's remember about abstraction) And better move this filed to the method where we use it
```suggestion
    private Predicate<Candidate> candidateValidator = new CandidateValidator();
```

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,74 @@
...
+
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(string -> Arrays.stream(string.split(NUMBER_SEPARATOR)))
+                .mapToInt(Integer::parseInt)
+                .filter(integer -> integer % 2 == 0)
+                .min()
+                .orElseThrow(() -> new RuntimeException("Can't get min value from list:"
```
Sofasmile, line 24:
```suggestion
                .orElseThrow(() -> new RuntimeException("Can't get min value from list:" + numbers
```

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,74 @@
...
-     * return the average of all odd numbers from the list or throw NoSuchElementException.
-     * But before that subtract 1 from each element on an odd position (having the odd index).
-     */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(index -> index % 2 != 0 ? numbers.get(index) - DECREMENT : numbers.get(index))
+                .filter(number -> number % 2 != 0)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
Sofasmile, line 33: it's optional to pass `NoSuchElementException` in the constructor of the method `orElseThrow`. It will be thrown by default

Sofasmile (changes requested):
great)

Sofasmile (approved):

PR #125, opened 2021-08-13
Solution TASK "stream practise"

Chupiak:


RomanDubovskyi (approved):
all good, I like it 👍

PR #135, opened 2021-09-05
stream practice solution

kryzhen:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,11 @@
...
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    @Override
+    public boolean test(Candidate candidate) {
+        return false;
```
MaliukDaria, line 9: We need to implement here candidate validation logic: "person should be older than 35 y, should be allowed to vote, have nationality - 'Ukrainian' and live in urk for 10 years. For the last requirement use field periodsInUkr, which has following view: "2002-2015" For now we don't care if that was last 10 or not."

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +37,11 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        OptionalDouble average = IntStream.range(0, numbers.size())
+                .map(i -> i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i))
+                .filter(i -> i % 2 != 0)
+                .average();
+        return average.orElseThrow(NoSuchElementException::new);
```
MaliukDaria, line 44:
```suggestion
return IntStream.range(0, numbers.size())
                .map(i -> i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i))
                .filter(i -> i % 2 != 0)
                .average()
                .getAsDouble();
```

we can make it simpler

src/main/java/practice/StreamPractice.java
```diff
@@ -34,8 +52,12 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * <p>
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
-    public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+    public List<Person> selectMenByAge(List<Person> peopleList,
+                                       int fromAge, int toAge) {
+        return peopleList.stream()
+                .filter(i -> (i.getSex() == Person.Sex.MAN) && (i.getAge() > fromAge)
```
MaliukDaria, line 58: We have to include `fromAge` ("men whose age is from `fromAge` to `toAge` inclusively")

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +72,11 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(i -> (i.getSex() == Person.Sex.MAN && i.getAge() <= maleToAge)
+                        || (i.getSex() == Person.Sex.WOMAN && i.getAge() <= femaleToAge))
+                .filter(i -> i.getAge() >= fromAge)
```
MaliukDaria, line 78: can we use only one `filter` operation here?

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +85,11 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(i -> i.getSex() == Person.Sex.WOMAN && i.getAge() >= femaleAge)
+                .flatMap(i -> i.getCats().stream())
```
MaliukDaria, line 90:
```suggestion
                .filter(person -> person.getSex() == Person.Sex.WOMAN && person.getAge() >= femaleAge)
                .flatMap(person -> person.getCats().stream())
```

I think it's more readable this way

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +105,14 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(i -> i.getAge() >= 35 && i.getNationality().equals("Ukrainian")
+                        && i.isAllowedToVote())
+                .filter(i -> Arrays.stream(i.getPeriodsInUkr().split("-"))
+                        .map(Integer::valueOf)
+                        .reduce(0, (a, b) -> b - a) > 10)
```
MaliukDaria, line 113: Let's implement and use `CandidateValidator` here

MaliukDaria (changes requested):

MaliukDaria (approved):
👍🏻

PR #136, opened 2021-09-06
DmytroPetrushenko / jv-stream-github-practice

DmytroPetrushenko:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,19 @@
 package practice;
 
-public class CandidateValidator {
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
     //write your code here
+    static final int validAge = 35;
+    static final int validLiveInUkraine = 10;
```
MaliukDaria, line 9: What about constant naming and access modifiers? https://mate-academy.github.io/style-guides/java/java.html#s5.2.4-constant-names

MaliukDaria (changes requested):

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,19 @@
 package practice;
 
-public class CandidateValidator {
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
     //write your code here
+    static final int VALID_AGE = 35;
+    static final int VALID_LIVE_IN_UKRAINE = 10;
```
MaliukDaria, line 9:
```suggestion
    private static final int VALID_AGE = 35;
    private static final int VALID_LIVE_IN_UKRAINE = 10;
```

don't forget about access modifiers

MaliukDaria (approved):

PR #140, opened 2021-09-14
Implemented java-stream task.

KyryloDvornyk:


romuth (approved):

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,22 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MIN_YEAR = 35;
```
MaliukDaria, line 7:
```suggestion
    private static final int MIN_AGE = 35;
```

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,22 @@
...
+    private static final int MIN_YEAR_UKRAINE = 10;
```
```suggestion
    private static final int MIN_YEARS_IN_UKRAINE = 10;
```

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,71 @@
...
-     * Example: select people of working age
-     * (from 18 y.o. and to 60 y.o. for men and to 55 y.o. for women inclusively).
-     */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        Predicate<Person> personPredicate = person -> (person.getAge() >= fromAge
+                && person.getAge() <= maleToAge && person.getSex() == Person.Sex.MAN)
+                || (person.getAge() >= fromAge && person.getAge() <= femaleToAge
+                && person.getSex() == Person.Sex.WOMAN);
```
MaliukDaria, line 48: could we make this check `person.getAge() >= fromAge` only ones?

MaliukDaria (changes requested):

MaliukDaria (approved):
🔥

PR #142, opened 2021-09-14
Hw 11 solution

7777vova7777:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,14 @@
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+
+        return numbers.stream()
+                .flatMap(s -> Arrays.stream(s.split(",")))
+                .map(Integer::valueOf)//.mapToInt(Integer::parseInt)
```
RomanDubovskyi, line 24: don't push commented code to remote repository

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +34,12 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+
+        return IntStream.range(0, numbers.size())
+                .map(i -> i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i))
+                .filter(i -> i % 2 != 0)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
RomanDubovskyi, line 42:
```suggestion
                .getAsDouble();
```
it will throw this exception by default if no value is present

RomanDubovskyi (changes requested):
good job, left 2 minor comments 👍

src/main/java/practice/StreamPractice.java
7777vova7777: Ok

src/main/java/practice/StreamPractice.java
7777vova7777: noted

RomanDubovskyi (approved):

PR #155, opened 2021-09-16
added solution

kvoytikh:
added solution


KyryloDvornyk (approved):

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,115 @@
...
+     * Given a List of Integer numbers,
+     * return the average of all odd numbers from the list or throw NoSuchElementException.
+     * But before that subtract 1 from each element on an odd position (having the odd index).
+     */
+    public Double getOddNumsAverage(List<Integer> numbers) {
+        return IntStream.range(0, numbers.size())
+                .map(index -> index % 2 == 1 ? numbers.get(index) - 1 : numbers.get(index))
+                .filter(number -> number % 2 == 1)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
MaliukDaria, line 43: what the difference between orElseThrow(Supplier<? extends X> exceptionSupplier) and orElseThrow()

MaliukDaria (changes requested):

src/main/java/practice/StreamPractice.java
kvoytikh: The orElseThrow(Supplier) method of OptionalInt class used to get the value contained by OptionalInt. If a value is present, this method returns the value, otherwise, this method throws an exception produced by the exception supplying function. 
orElseTrow() - returns the value / throws an exception

MaliukDaria (approved):
🔥

PR #158, opened 2021-09-16
implemented tasks with streams

SergiiLonskyi:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,20 @@
...
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final String NATIONALITY = "Ukrainian";
+    private static final int MIN_AGE = 35;
+    private static final int LIVE_IN_UKRAINE = 10;
+
+    @Override
+    public boolean test(Candidate candidate) {
+        int yearFrom = Integer.parseInt(candidate.getPeriodsInUkr().split("-")[0]);
+        int yearTo = Integer.parseInt(candidate.getPeriodsInUkr().split("-")[1]);
```
RostyslavF5R, line 14: Don't use magic numbers, create constant fo 1 and 0

RostyslavF5R (changes requested):

RostyslavF5R (approved):

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,24 @@
...
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final String NATIONALITY = "Ukrainian";
+    private static final int MIN_AGE = 35;
+    private static final int LIVE_IN_UKRAINE = 10;
```
Max-Pochepets, line 9:
```suggestion
    private static final int REQUIRED_TIME_IN_UKRAINE = 10;
```

Max-Pochepets (approved):

PR #168, opened 2021-09-16
First solution

taras-oleksiuk:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,21 @@
...
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int CANDIDATE_REQUIRED_AGE = 35;
+    private static final int CANDIDATE_LIVE_IN_UKR = 10;
+    private static final String CANDIDATE_NATIONALITY = "Ukrainian";
+
+    @Override
+    public boolean test(Candidate candidate) {
+        int livingPeriod = Integer.parseInt(candidate.getPeriodsInUkr().substring(5))
+                - Integer.parseInt(candidate.getPeriodsInUkr().substring(0, 4));
```
StasShepet, line 14: 0, 4, 5 - must be constant

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +50,11 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        Predicate<Person> personFilter = person -> person.getSex().equals(Person.Sex.MAN)
+                && person.getAge() >= fromAge && person.getAge() <= toAge;
```
StasShepet, line 54: maybe predicate here is redundant

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +69,14 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        Predicate<Person> personFilter = person -> person.getAge() >= fromAge
+                && ((person.getSex().equals(Person.Sex.MAN)
+                && person.getAge() <= maleToAge)
+                || person.getSex().equals(Person.Sex.WOMAN)
+                && person.getAge() <= femaleToAge);
```
StasShepet, line 76: and here

StasShepet (changes requested):

src/main/java/practice/CandidateValidator.java
taras-oleksiuk: I don't think so, they don't use often

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,13 @@
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
```
StasShepet, line 19: let's delete all coments

StasShepet (approved):

Max-Pochepets (approved):

PR #177, opened 2021-09-25
Home work

Noct2000:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,18 @@
...
+    @Override
+    public boolean test(Candidate candidate) {
+        int[] periodsInUkr = Arrays.stream(candidate.getPeriodsInUkr().split("-"))
+                .mapToInt(Integer::parseInt).toArray();
+        boolean hasTenYearsInUkr = periodsInUkr[1] - periodsInUkr[0] >= 10;
+        boolean isUkrainian = candidate.getNationality().equals("Ukrainian");
+        boolean allowedAge = candidate.getAge() >= 35;
+        return hasTenYearsInUkr && candidate.isAllowedToVote() && isUkrainian && allowedAge;
+    }
     //write your code here
```
RomanDubovskyi, line 17: let's remove comment

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +33,16 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(i -> {
+                    if (i % 2 != 0) {
+                        return numbers.get(i) - 1;
```
RomanDubovskyi, line 39: let's use ternary operator here, it will shorten the code

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +33,16 @@ public int findMinEvenNumber(List<String> numbers) {
...
+                    } else {
+                        return numbers.get(i);
+                    }
+                })
+                .filter(n -> n % 2 != 0)
+                .average().orElseThrow(NoSuchElementException::new);
```
RomanDubovskyi, line 45:
```suggestion
                .average()
                .getAsDouble()
```
it will generate this exception by default if there is no value

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +72,17 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        Predicate<Person> selectMenPredicate = p -> p.getSex() == Person.Sex.MAN
```
RomanDubovskyi, line 75: we don't need to use predicate object all the time. If logical construction is not reused or complex let's put logic construction directly into filter() operation

RomanDubovskyi (changes requested):

RomanDubovskyi (approved):

PR #178, opened 2021-09-27
solution made all tests passed

ikhominitch:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,20 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int FIRST_DATE = 0;
```
RomanDubovskyi, line 7:
```suggestion
    private static final int START_DATE_INDEX = 0;
    private static final int END_DATE_INDEX = 1;
```

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +36,11 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(i -> i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i))
+                .filter(i -> i % 2 != 0)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
RomanDubovskyi, line 43:
```suggestion
                .getAsDouble()
```
this method generates NoSuchElementExceptioin by default if there is no value

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +52,13 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        Predicate<Person> menByAge = person -> person.getSex() == Person.Sex.MAN
```
RomanDubovskyi, line 55: If you don't reuse logic construction and it's not big you can put it inside of filter operation right away

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +73,18 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        Predicate<Person> workablePeople = new Predicate<>() {
```
RomanDubovskyi, line 76: same here, we don't need to craete predicate object in each use case

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +93,12 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> femaleAge <= p.getAge() && p.getSex() == Person.Sex.WOMAN)
+                .map(Person::getCats)
+                .flatMap(Collection::stream)
```
RomanDubovskyi, line 99:
```suggestion
                .flatMap(p -> p.getCats().stream())
```

RomanDubovskyi (changes requested):
good job, left minor comments

RomanDubovskyi (approved):

PR #182, opened 2021-11-04
implemented task stream

alex-hazniuk:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,28 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int FIRST_INDEX = 0;
```
RomanDubovskyi, line 7:
```suggestion
    private static final int BEGIN_YEAR_INDEX = 0;
```
it's better to name variables according to their purpose not the value that hold

RomanDubovskyi (approved):
all good, left one minor comment

PR #186, opened 2021-11-12
all methods are implemented

Andrey-Sidorkin:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,14 @@
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(s -> s.split(","))
+                .flatMap(Arrays::stream)
```
QQBratka, line 25:
```suggestion
                .flatMap(n -> Arrays.stream(n.split(",")))
```you can do this and save one line life))

File: src/main/java/practice/StreamPractice.java [hunk 2]
```diff
@@ -59,22 +86,33 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        Predicate<Person> personPredicate = person -> person.getSex() == Person.Sex.WOMAN
+                && person.getAge() >= femaleAge;
+        return peopleList.stream()
+                .filter(personPredicate)
+                .map(Person::getCats)
+                .flatMap(Collection::stream)
```
QQBratka, line 94:
```suggestion
                .flatMap(el -> el.getCats().stream())
```another line want to live))

**Review by QQBratka — CHANGES_REQUESTED** [review]

File: src/main/java/practice/StreamPractice.java [hunk 3]
```diff
@@ -23,7 +36,12 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .mapToDouble(i -> numbers.get(i) - i % 2)
+                .filter(n -> n % 2 != 0)
+                .average()
+                .orElseThrow(() -> new NoSuchElementException(
+                        "Can't get average from list: " + numbers));
```
QQBratka, line 44: If you need to throw exception when no result found after stream execution sometimes get()/getAsDouble() may help, try to google what does it do
I do same mistake and after read try to avoid) 
```suggestion
                .getAsDouble();
```

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,19 @@
...
+    public static final int SUITABLE_PERIOD_IN_UKRAINE = 10;
+
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.getNationality().equals("Ukrainian")
+            && Integer.parseInt(candidate.getPeriodsInUkr()
+                .replaceAll("\\d+-", ""))
+            - Integer.parseInt(candidate.getPeriodsInUkr()
+                .replaceAll("-\\d+", "")) >= SUITABLE_PERIOD_IN_UKRAINE
+            && candidate.getAge() >= SUITABLE_AGE && candidate.isAllowedToVote();
```
QQBratka, line 17: create field for nationality, imagin that this code will be need for voting in Trinidad & Tobago, but thair rules allows to be president only for who:
SUITABLE_AGE = 47;
SUITABLE_PERIOD_IN_COUNTRY = 18;
NATIONALITY = "Trinidadian";
Actually about that thing were written in try to avoid common mistakes) 
And also for this for better reading use methods fields like:
int startLivePeriod = Integer.parseInt(candidate.getPeriodsInUkr()
                .replaceAll("\\d+-", ""))
int endLeavePeriod = Integer.parseInt(candidate.getPeriodsInUkr()
                .replaceAll("-\\d+", ""))

QQBratka (approved):

Sofasmile (approved):

PR #201, opened 2021-11-18
implemented class CandidateValidator and added 6 filter statistics me…

PineappleApplePinecone:
…thods

### DIFF
--- a/src/main/java/practice/CandidateValidator.java
+++ b/src/main/java/practice/CandidateValidator.java
@@ -1,5 +1,23 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MINIMUM_AGE = 35;
+    private static final String NATIONALITY = "Ukrainian";
+    private static final int LIVING_IN_UKRAINE = 10;
+    private static final int FIRST_DATE_INDEX = 0;
+    private static final int LAST_DATE_INDEX = 1;
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] period = candidate.getPeriodsInUkr().split("-");
+        int fromYear = Integer.parseInt(period[FIRST_DATE_INDEX]);
+        int toYear = Integer.parseInt(period[LAST_DATE_INDEX]);
+        return candidate.getAge() >= MINIMUM_AGE
+                && candidate.getNationality().equals(NATIONALITY)
+                && toYear - fromYear >= LIVING_IN_UKRAINE
+                && candidate.isAllowedToVote();
+    }
 }
--- a/src/main/java/practice/StreamPractice.java
+++ b/src/main/java/practice/StreamPractice.java
@@ -1,8 +1,12 @@
 package practice;
 
-import java.util.Collections;
+import java.util.Arrays;
 import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import model.Candidate;
+import model.Cat;
 import model.Person;
 
 public class StreamPractice {
@@ -14,7 +18,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(y -> Arrays.stream(y.split(",")))
+                .mapToInt(Integer::parseInt)
+                .filter(y -> y % 2 == 0)
+                .min()
+                .orElseThrow(() ->
+                        new RuntimeException("Can't get min value from list: " + numbers));
     }
 
     /**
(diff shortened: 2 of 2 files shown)


src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +83,13 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList
+                .stream()
+                .filter(p -> p.getSex().equals(Person.Sex.WOMAN)
+                        && p.getAge() >= femaleAge && !p.getCats().isEmpty())
```
KostyaNikiforov, line 89: I think it would be better to use here an implementation of Predicate

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +68,13 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getAge() >= fromAge
+                        && (p.getSex() == Person.Sex.MAN
+                        && p.getAge() <= maleToAge
+                        || p.getSex() == Person.Sex.WOMAN
+                        && p.getAge() <= femaleToAge))
```
KostyaNikiforov, line 76: Better use Predicate

KostyaNikiforov: We have this point in the common mistake list

KostyaNikiforov (changes requested):

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +49,11 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex() == Person.Sex.MAN
+                        && p.getAge() >= fromAge && p.getAge() <= toAge)
```
KostyaNikiforov, line 54: The same

KostyaNikiforov (changes requested):

src/main/java/practice/StreamPractice.java
PineappleApplePinecone: Okay got it, look again, when you have got free time

KostyaNikiforov (approved):

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,23 @@
...
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MINIMUM_AGE = 35;
+    private static final String NATIONALITY = "Ukrainian";
+    private static final int LIVING_IN_UKRAINE = 10;
+    private static final int FIRST_DATE_INDEX = 0;
+    private static final int LAST_DATE_INDEX = 1;
```
Nick97-git, line 11:
```suggestion
    private static final int MINIMUM_AGE = 35;
    private static final String REQUIRED_NATIONALITY = "Ukrainian";
    private static final int MIN_YEARS_TO_LIVE_IN_UKRAINE = 10;
    private static final int YEAR_FROM_INDEX = 0;
    private static final int YEAR_TO_INDEX = 1;
```

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,13 @@
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(y -> Arrays.stream(y.split(",")))
```
Nick97-git, line 23: think about better naming, you have a collection of numbers, so how do you think you could name it?
the same with other similar situations

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +85,14 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        Predicate<Person> personPredicate = p -> p.getSex().equals(Person.Sex.WOMAN)
```
Nick97-git, line 88: how we should compare two enums?

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +85,14 @@ public Double getOddNumsAverage(List<Integer> numbers) {
...
+                && p.getAge() >= femaleAge && !p.getCats().isEmpty();
```
Nick97-git, line 89: do we need the last check?

Nick97-git (changes requested):

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,13 @@
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(number -> Arrays.stream(number.split(",")))
+                .mapToInt(Integer::parseInt)
+                .filter(y -> y % 2 == 0)
```
Nick97-git, line 25:
```suggestion
                .filter(number -> number % 2 == 0)
```

Nick97-git (approved):

PR #213, opened 2021-11-19
Implemented task

DanyNest:


Anatolii-Huryk (approved):

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +69,14 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(e -> (e.getSex() == Person.Sex.MAN
+                        && e.getAge() >= fromAge
```
Sofasmile, line 74: we have check `e.getAge() >= fromAge` twice, let'd do it only one time

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +106,10 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        Predicate<Candidate> candidateValidator = new CandidateValidator();
+        return candidates.stream().filter(candidateValidator)
```
Sofasmile, line 110:
```suggestion
        return candidates.stream()
                .filter(candidateValidator)
```

Sofasmile (changes requested):

Sofasmile (approved):

PR #229, opened 2021-12-23
Finished Stream Practice. Implements own Predicate

Viktor-Stadnik:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,17 @@
...
+
+public class CandidateValidator implements Predicate<Candidate> {
+    @Override
+    public boolean test(Candidate candidate) {
+        int yearTo = Integer.parseInt(candidate.getPeriodsInUkr()
+                .substring(0, candidate.getPeriodsInUkr().indexOf("-")));
+        int yearAfter = Integer.parseInt(candidate.getPeriodsInUkr()
+                .substring(candidate.getPeriodsInUkr().indexOf("-") + 1));
+        return (yearAfter - yearTo) > 10 && candidate.getAge() >= 35
+                && candidate.getNationality().equals("Ukrainian")
```
MaliukDaria, line 14: let's make 10, 35 and "Ukrainian" constant fields

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,65 @@
...
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(s -> s.split(","))
+                .flatMap(Arrays::stream)
+                .mapToInt(f -> Integer.parseInt(f))
+                .filter(n -> n % 2 == 0)
+                .min()
+                .orElseThrow(() -> new RuntimeException("Can't get min value from list: "
+                        + numbers));
```
MaliukDaria, line 21:
```suggestion
        return numbers.stream()
		.flatMap(n -> Arrays.stream(n.split(",")))
                .mapToInt(f -> Integer.parseInt(f))
                .filter(n -> n % 2 == 0)
                .min()
                .orElseThrow(() -> new RuntimeException("Can't get min value from list: "
                        + numbers));
```

we can combine map and flatMap

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,65 @@
...
-     */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> fromAge <= p.getAge() && p.getAge() <= maleToAge
+                        && p.getSex() == Person.Sex.MAN
+                        || fromAge <= p.getAge() && p.getAge() <= femaleToAge
+                        && p.getSex() == Person.Sex.WOMAN)
+                .collect(Collectors.toList());
```
MaliukDaria, line 47: could we use `p -> fromAge` only ones here?

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #230, opened 2021-12-24
realized test method from Predicate interface and other methods

ev-lanovenko:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,14 @@
...
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        int i = numbers.stream()
+                .flatMap(n -> Arrays.stream(n.split(",")))
+                .mapToInt(Integer::parseInt)
+                .filter(n -> n % 2 == 0)
+                .min()
+                .orElseThrow(() ->
+                        new RuntimeException("Can't get min value from list: " + numbers));
+        return i;
```
MaliukDaria, line 29:
```suggestion
        return numbers.stream()
                .flatMap(n -> Arrays.stream(n.split(",")))
                .mapToInt(Integer::parseInt)
                .filter(n -> n % 2 == 0)
                .min()
                .orElseThrow(() ->
                        new RuntimeException("Can't get min value from list: " + numbers));
```

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +35,12 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        double asDouble = IntStream.range(0, numbers.size())
+                .mapToDouble(i -> numbers.get(i) - i % 2)
+                .filter(i -> i % 2 != 0)
+                .average()
+                .getAsDouble();
+        return asDouble;
```
MaliukDaria, line 43:
```suggestion
        return IntStream.range(0, numbers.size())
                .mapToDouble(i -> numbers.get(i) - i % 2)
                .filter(i -> i % 2 != 0)
                .average()
                .getAsDouble();
```

and so on...

MaliukDaria (changes requested):

MaliukDaria (approved):
👍🏻

PR #236, opened 2022-01-30
My homework jv-stream-github-practice was done

roman-hrechanyi:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,23 @@
...
-public class CandidateValidator {
-    //write your code here
+import java.util.Arrays;
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MinAge = 35;
+    private static final int MinLivingAge = 10;
+    private static final String Nationality = "Ukrainian";
```
MaliukDaria, line 10: what about constant naming https://mate-academy.github.io/style-guides/java/java.html#s5.2.4-constant-names ?

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +17,17 @@
...
+            return numbers.stream()
+                    .map(w -> w.split(","))
+                    .flatMap(Arrays::stream)
+                    .mapToInt(Integer::parseInt)
+                    .filter(i -> i % 2 == 0)
+                    .min()
+                    .orElseThrow();
+        } catch (RuntimeException e) {
+            throw new RuntimeException("Can't get min value from list");
+        }
```
```suggestion
            return numbers.stream()
                    .flatMap(n -> Arrays.stream(n.split(",")))
                    .mapToInt(Integer::parseInt)
                    .filter(i -> i % 2 == 0)
                    .min()
                    .orElseThrow( () -> throw new RuntimeException("Can't get min value from list"););
```

why do we need try-catch block here? We can combine map and flatMap in one flatMap

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +36,16 @@ public int findMinEvenNumber(List<String> numbers) {
...
+        for (int i = 0; i < numbers.size(); i++) {
+            if (i % 2 != 0) {
+                numbers.set(i, numbers.get(i) - 1);
+            }
+        }
+        return numbers.stream()
+                .filter(i -> i % 2 != 0)
+                .mapToInt(Integer::intValue)
+                .average()
+                .orElseThrow();
```
MaliukDaria, line 48: lets do this only with Stream API (without loops)

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +92,12 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex().equals(Person.Sex.WOMAN) && p.getAge() >= femaleAge)
+                .map(Person::getCats)
+                .flatMap(Collection::stream)
```
MaliukDaria, line 98: cam we combine map and flatMap here?

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #239, opened 2022-02-03
first ver

LinkClink:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,24 @@
...
+    @Override
+    public boolean test(Candidate candidate) {
+        int[] periods = Arrays.stream(candidate.getPeriodsInUkr().split("-"))
+                .mapToInt(Integer::parseInt)
+                .toArray();
+        return candidate.isAllowedToVote()
+                && NATIONALITY.equals(candidate.getNationality())
+                && candidate.getAge() >= MIN_AGE
+                && periods[1]
+                - periods[0]
```
sviat-h, line 21: maybe we should also create constants for indexes, but i'm not sure

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,61 @@
...
-     * return the average of all odd numbers from the list or throw NoSuchElementException.
-     * But before that subtract 1 from each element on an odd position (having the odd index).
-     */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(e -> e % 2 != 0 ? numbers.get(e) - 1 : numbers.get(e))
+                .filter(e -> e % 2 != 0)
+                .average()
+                .getAsDouble();
```
sviat-h, line 27: we should return the average or throw NoSuchElementException here
```suggestion
                .orElseThrow(NoSuchElementException::new);
```

sviat-h (approved):

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,62 @@
...
-     * return min integer value. One more thing - we're interested in even numbers.
-     * If there is no needed data throw RuntimeException with message
-     * "Can't get min value from list: < Here is our input 'numbers' >"
-     */
+
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(e -> Arrays.stream(e.split(",")))
+                .map(Integer::parseInt).filter(integer -> integer % 2 == 0)
```
KostyaNikiforov, line 17:
```suggestion
                .map(Integer::parseInt)
                .filter(integer -> integer % 2 == 0)
```

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,62 @@
...
-     * return the average of all odd numbers from the list or throw NoSuchElementException.
-     * But before that subtract 1 from each element on an odd position (having the odd index).
-     */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(e -> e % 2 != 0 ? numbers.get(e) - 1 : numbers.get(e))
+                .filter(e -> e % 2 != 0)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
KostyaNikiforov, line 28: we have method that throws this exception by default

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,62 @@
...
-     * select from the List only men whose age is from `fromAge` to `toAge` inclusively.
-     * <p>
-     * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
-     */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(e -> e.getSex() == Person.Sex.MAN
+                        && e.getAge() >= fromAge && e.getAge() <= toAge)
+                .collect(Collectors.toList());
```
KostyaNikiforov, line 35: It is Ok, but to make it more clearly, you can move large conditions to a separate Predicate function

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,62 @@
...
-     * (from 18 y.o. and to 60 y.o. for men and to 55 y.o. for women inclusively).
-     */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(e -> e.getAge() >= fromAge
+                        && ((e.getSex() == Person.Sex.MAN && e.getAge() <= maleToAge)
+                        || (e.getSex() == Person.Sex.WOMAN && e.getAge() <= femaleToAge)))
+                .collect(Collectors.toList());
```
KostyaNikiforov, line 44: The same here

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):

PR #244, opened 2022-02-05
created all methods & class CandidateValidator

yurii-tomashevskyi:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,14 @@
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers
+                .stream()
```
liliia-ponomarenko, line 23: not need separate stream from numbers

liliia-ponomarenko (approved):

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,21 @@
...
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MINIMAL_AGE = 35;
+    private static final String NATIONALITY = "Ukrainian";
+    private static final int MINIMAL_LIVE_IN_UKRAINE = 10;
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] yearsLiveInUkraine = candidate.getPeriodsInUkr().split("-");
+        int numberYearsLiveInUkraine = Integer.parseInt(yearsLiveInUkraine[1])
+                - Integer.parseInt(yearsLiveInUkraine[0]);
```
KostyaNikiforov, line 15: Better make those magic numbers constant too

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +71,13 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList
+                .stream()
+                .filter(p -> (p.getSex().equals(Person.Sex.MAN)
+                        && p.getAge() >= fromAge && p.getAge() <= maleToAge)
+                        || (p.getSex().equals(Person.Sex.WOMAN)
+                        && p.getAge() >= fromAge && p.getAge() <= femaleToAge))
```
KostyaNikiforov, line 79: you can move this condition to the predicate function too, but it is up to you

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +108,11 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates
+                .stream()
+                .filter(c -> new CandidateValidator().test(c))
```
KostyaNikiforov, line 113:
```suggestion
                .filter(new CandidateValidator())
```

KostyaNikiforov (changes requested):

PR #246, opened 2022-02-07
Solution of the task

k1llzers:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +17,17 @@
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        if (numbers.size() == 0) {
+            throw new RuntimeException("Can't get min value from list: " + numbers);
+        }
+        return numbers.stream()
+                .map(separation -> separation.split(","))
+                .flatMap(transition -> Arrays.stream(transition))
```
RomanDubovskyi, line 25:
```suggestion
                .flatMap(nums -> Arrays.stream(nums.split(",")))
```
we can combine these 2 operations into one

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +17,17 @@
...
+                .mapToInt(transition -> Integer.parseInt(String.valueOf(transition)))
+                .filter(parity -> parity % 2 == 0)
+                .reduce(Integer::min)
+                .getAsInt();
```
RomanDubovskyi, line 29:
```suggestion
                .min(Integer::compareTo)
                .orElseThrow...
```

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +17,17 @@
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        if (numbers.size() == 0) {
```
RomanDubovskyi, line 20: this can be done at the end of the stream

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +36,26 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
```
RomanDubovskyi, line 38: you should use only Stream here, try to apply the Instream.range() operation to get stream of indecies. Then map them into numbers depending on the index, if it's odd or even

RomanDubovskyi (changes requested):

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +33,31 @@ public int findMinEvenNumber(List<String> numbers) {
...
+                        return numbers.get(number);
+                    }
+                })
+                .filter(odd -> odd % 2 != 0)
+                .reduce(Integer::sum)
+                .get();
+        if (count == 0) {
+            throw new NoSuchElementException("No such odd element");
+        }
+        return (double) sum / (double) count;
```
Resci, line 60:
```suggestion
return IntStream.range(0, numbers.size())
    .map(i -> i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i))
    .filter(i -> i % 2 != 0)
    .average()
    .getAsDouble();
```

Resci (approved):

Resci (approved):

PR #255, opened 2022-02-13
Homework

karma-o:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,90 @@
...
+    public static final String COMMA_REGEX = ",";
+
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        Predicate<Integer> evenPredicate = i -> i % 2 == 0;
+        return numbers.stream()
+                .flatMap(s -> Arrays.stream(s.split(COMMA_REGEX)))
+                .map(Integer::parseInt)
+                .filter(evenPredicate)
+                .mapToInt(Integer::valueOf)
```
DaniilOpryshko, line 22: Do we really need this line?

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,90 @@
...
+        return numbers.stream()
+                .map(n -> {
+                    if (index.get() % 2 != 0) {
+                        n = n - 1;
+                    }
+                    index.getAndIncrement();
+                    return n;
+                })
+                .filter(oddNumberPredicate)
+                .mapToDouble(Integer::doubleValue)
```
DaniilOpryshko, line 40: We don't need this line

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,90 @@
...
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
DaniilOpryshko, line 42: If you need to throw exception when no result found after stream execution sometimes get()/getAsDouble() may help, try to google what does it do

DaniilOpryshko (changes requested):

src/main/java/practice/StreamPractice.java
karma-o: Once again we need double to use `avarage()`, not `Integer`

src/main/java/practice/StreamPractice.java
karma-o: Yes, because we need a stream of primitives to perform a `sum()` method, `Integer` doesn`t fit here

DaniilOpryshko: We can use method min with Stream<Integer>, we just need to pass Comparator in it

DaniilOpryshko (approved):

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,88 @@
...
+        AtomicInteger index = new AtomicInteger();
+        Predicate<Integer> oddNumberPredicate = n -> n % 2 != 0;
+        return numbers.stream()
+                .map(n -> {
+                    if (index.get() % 2 != 0) {
+                        n = n - 1;
+                    }
+                    index.getAndIncrement();
+                    return n;
+                })
```
AlexRozv, line 36: to simplify this solution we can do something like this
IntStream.range(0, numbers.size()).map(n -> n % 2 == 0 ? numbers.get(n) : numbers.get(n) - 1)

AlexRozv (changes requested):

AlexRozv (approved):
good job 👍

PR #268, opened 2022-02-15
1st implementation of jv-stream-github-practice task

OleksiyTitov:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,74 @@
...
-     * But before that subtract 1 from each element on an odd position (having the odd index).
-     */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(i -> i % 2 == 1 ? (numbers.get(i) - 1) : numbers.get(i))
+                .filter(n -> n % 2 == 1)
+                .average()
+                .getAsDouble();
     }
```
sltnsrh, line 34: .average() already returns Double,  .getAsDouble() may be not necessary.
* return the average of all odd numbers from the list or throw NoSuchElementException. Let's add exception as in condition.

sltnsrh (changes requested):

sltnsrh (approved):

AlexRozv (approved):
great job 🔥

PR #270, opened 2022-02-16
implemented all methods and class CandidateValidator

antonignatenko:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,26 @@
...
+    private static final int MIN_YEARS_IN_UKRAINE = 10;
+
+    @Override
+    public boolean test(Candidate candidate) {
+
+        return candidate.getAge() >= MIN_ELIGIBLE_AGE
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals(NATIONALITY)
+                && isEligible(candidate.getPeriodsInUkr());
+
```
Sasha-Bachynskyi, line 18: and here

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,26 @@
...
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MIN_ELIGIBLE_AGE = 35;
+    private static final String NATIONALITY = "Ukrainian";
+    private static final int MIN_YEARS_IN_UKRAINE = 10;
+
+    @Override
+    public boolean test(Candidate candidate) {
+
```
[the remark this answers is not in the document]
AlexRozv: agree

src/main/java/practice/CandidateValidator.java
AlexRozv: agree

src/main/java/practice/CandidateValidator.java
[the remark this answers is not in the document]
antonignatenko: can't avoid this line
mvn won't approve

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +83,12 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex().equals(Person.Sex.WOMAN) && p.getAge() >= femaleAge)
+                .map(person -> person.getCats())
+                .flatMap(cats -> cats.stream())
```
AlexRozv, line 89: instead of using map and then flatmap we can just use one flatmap here, try combining it

AlexRozv (changes requested):

AlexRozv (approved):
✅

PR #278, opened 2022-03-19
solution completed

Hryniv:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,19 @@
...
-        return 0;
+        return numbers
+                .stream()
+                .map(e -> List.of(e.split(",")))
+                .flatMap(Collection::stream)
+                .map(Integer::valueOf)
+                .mapToInt(e -> e)
+                .filter(e -> e % 2 == 0)
+                .min()
+                .orElseThrow(() -> new RuntimeException(massage(numbers)));
```
```suggestion
        return numbers.stream()
                .flatMap(n -> Arrays.stream(n.split(",")))
                .map(Integer::valueOf)
                .filter(e -> e % 2 == 0)
                .min()
                .orElseThrow(() -> new RuntimeException(massage(numbers)));
```

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +40,11 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(i -> i % 2 == 1 ? numbers.get(i) - 1 : numbers.get(i))
+                .filter(i -> i % 2 == 1)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
MaliukDaria, line 47:
```suggestion
                .orElseThrow();
```

What `orElseThrow` method throws by default?

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,19 @@
...
+                .map(e -> List.of(e.split(",")))
+                .flatMap(Collection::stream)
+                .map(Integer::valueOf)
+                .mapToInt(e -> e)
+                .filter(e -> e % 2 == 0)
+                .min()
+                .orElseThrow(() -> new RuntimeException(massage(numbers)));
+    }
+
+    private static String massage(List<String> numbers) {
```
MaliukDaria, line 33: we use this method only ones, i dont think we need to create separate method for this logic

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +92,14 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList
+                .stream()
+                .filter(i -> i.getSex() == Person.Sex.WOMAN && i.getAge() >= femaleAge)
+                .map(Person::getCats)
+                .flatMap(Collection::stream)
```
MaliukDaria, line 99:
```suggestion
                .flatMap(p -> p.getCatList().stream())
```

lets combine map and flatMap here

MaliukDaria (changes requested):

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,28 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.Arrays;
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+
```
```suggestion
public class CandidateValidator implements Predicate<Candidate> {
    private static final int OPTIMAL_AGE = 35;
    private static final int VALID_YEARS_IN_UKR = 10;
    private static final String VALID_NATIONALITY = "Ukrainian";
```

I meant only `private static final boolean ALLOWED = true;` constant, we need the rest

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #288, opened 2022-04-19
feat: add implementation of methods using StreamAPI

mikenikolov:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,20 +1,37 @@
...
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(e -> e.split(","))
+                .flatMap(Arrays::stream)
+                .mapToInt(Integer::parseInt)
+                .filter(isOddNumber.negate())
+                .min()
+                .orElseThrow(
+                        () -> new RuntimeException("Can't get min value from list: " + numbers));
```
MaliukDaria, line 34:
```suggestion
                .orElseThrow(() -> new RuntimeException(
                		"Can't get min value from list: " + numbers));
```

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +84,11 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getAge() >= femaleAge && p.getSex() == Person.Sex.WOMAN)
```
MaliukDaria, line 88: let's also use a separate predicate here(from private method) for code consistency

MaliukDaria (approved):
👍🏻

PR #315, opened 2022-04-26
Implemented tasks

Syon2154:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,23 @@
...
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final String NATIONALITY = "Ukrainian";
+    private static final int MIN_AGE = 35;
+    private static final int MIN_YEARS_IN_UKRAINE = 10;
+    private static final int INDEX_OF_FIRST_YEAR = 0;
+    private static final int INDEX_OF_LAST_YEAR = 1;
```
Roman-Kachmaryk, line 11: Do we really need these two fields here? I don't think they are that important

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,23 @@
...
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] years = candidate.getPeriodsInUkr().split("-");
+        int period = Integer.parseInt(years[INDEX_OF_LAST_YEAR])
+                - Integer.parseInt(years[INDEX_OF_FIRST_YEAR]);
```
Roman-Kachmaryk, line 17: You can create a separate method to check if a candidate has spent at least 10 years in Ukraine

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,15 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers
+                .stream()
```
Roman-Kachmaryk, line 22:
```suggestion
        return numbers.stream()
```

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,14 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(s -> s.split(","))
+                .flatMap(Arrays::stream)
```
KseniiaMakarova, line 23: let's try to use only flatMap here

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +68,10 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        PersonValidator personValidator = new PersonValidator(fromAge, femaleToAge, maleToAge);
```
KseniiaMakarova, line 71: use abstract type for variable

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +100,11 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        CandidateValidator candidateValidator = new CandidateValidator();
```
KseniiaMakarova, line 103: same

src/test/java/core/basesyntax/StreamPracticeTest.java
```diff
@@ -1,11 +1,5 @@
 package core.basesyntax;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
-import java.util.NoSuchElementException;
-import java.util.function.Predicate;
```
KseniiaMakarova, line 8: please look into `Files changed` tab on Github and see what changes you made to the test class

this is not a big problem because you only moved methods and imports (and not changed the code of tests)
but it is better not to touch test class at all, so let's move everything back

KseniiaMakarova (changes requested):

src/test/java/core/basesyntax/StreamPracticeTest.java
Syon2154: It happened by mistake during formatting

PR #324, opened 2022-05-16
Stream solution added

PavloMykhalov:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,14 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(n -> n.split(","))
+                .flatMap(Arrays::stream)
```
MaliukDaria, line 23: could we combine map and flatMap to one flatMap?

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +34,12 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(n -> n % 2 != 0 ? numbers.get(n) - 1 : numbers.get(n))
+                .filter(n -> n % 2 != 0)
+                .mapToDouble(Double::valueOf)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
MaliukDaria, line 42: what orElseThrow() method throw by default?

MaliukDaria (changes requested):

PR #330, opened 2022-05-29
implemented required methods

viktoriia-haraniuk:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,26 @@
...
+        return candidate.getAge() >= MIN_ALLOWED_AGE
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals(REQUIRED_NATIONALITY)
+                && getPeriodOfLiving(candidate.getPeriodsInUkr()) >= PERIOD_OF_LIVING;
+    }
+
+    private int getPeriodOfLiving(String period) {
+        String[] dates = period.split("-");
+        return Integer.parseInt(dates[TO_PERIOD])
+                - Integer.parseInt(dates[FROM_PERIOD]);
```
Sm0kk, line 24: If you want you cant do this via stream.

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +51,12 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        Predicate<Person> personPredicate = person -> person.getSex().equals(Person.Sex.MAN)
```
Sm0kk, line 54: I think enums need to compare using the '==' operator, because equals can throw NPE.  And below also

Sm0kk (changes requested):

KostyaNikiforov (approved):
LGTM:fire:

PR #331, opened 2022-05-29
tests and stylechecks passed

drapail:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +21,18 @@ public class StreamPractice {
...
+        try {
+            i = Arrays.stream(String.join(",", numbers)
+                            .split(","))
+                    .map(Integer::parseInt)
+                    .filter(n -> n % 2 == 0)
+                    .mapToInt(n -> n)
+                    .min();
+        } catch (RuntimeException e) {
+            throw new RuntimeException("Can't get min value from list: " + numbers);
+        }
```
MaliukDaria, line 34: Can we make this without try-catch

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +41,18 @@ public int findMinEvenNumber(List<String> numbers) {
...
+                .map(numbers::get)
+                .map(n -> n - 1);
+        IntStream evenNumbers = IntStream.range(0, numbers.size())
+                .filter(i -> i % 2 == 0)
+                .map(i -> numbers.get(i));
+
+        return IntStream.concat(intStream, evenNumbers)
+                .filter(p -> p % 2 == 1)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
MaliukDaria, line 55: we need to use only one Stream here

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +64,11 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        List<Person> collect = peopleList.stream()
+                .filter(p -> p.getSex() == Person.Sex.MAN && p.getAge() > fromAge
+                        && p.getAge() <= toAge)
+                .collect(Collectors.toList());
+        return collect;
```
MaliukDaria, line 71:
```suggestion
        return peopleList.stream()
                .filter(p -> p.getSex() == Person.Sex.MAN 
                        && p.getAge() > fromAge
                        && p.getAge() <= toAge)
                .collect(Collectors.toList());
```

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +83,14 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
...
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+
+        List<Person> collect = peopleList.stream()
+                .filter(p -> ((p.getSex() == Person.Sex.MAN) && (p.getAge() >= fromAge)
+                        && (p.getAge() <= maleToAge))
+                        || ((p.getSex() == Person.Sex.WOMAN) && (p.getAge() >= fromAge)
+                        && (p.getAge() <= femaleToAge)))
+                .collect(Collectors.toList());
```
MaliukDaria, line 92: can we make `p.getAge() >= fromAge` check only ones here

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +99,13 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> ((p.getSex() == Person.Sex.WOMAN) && (p.getAge() >= femaleAge)))
+                .map(Person::getCats)
+                .flatMap(Collection::stream)
```
MaliukDaria, line 105: can we combine map and flatMap in one flatMap

MaliukDaria (changes requested):

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,20 @@
 package practice;
 
-public class CandidateValidator {
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    public static final String NATIONALITY = "Ukrainian";
+    public static final int MIN_AGE = 35;
+    public static final int TIME_IN_UKRAINE = 10;
```
MaliukDaria, line 9: we can make these constant private if we use them only in this class

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,16 @@ public class StreamPractice {
...
+        if (numbers.isEmpty()) {
+            throw new RuntimeException("Can't get min value from list: " + numbers);
+        }
+        OptionalInt i = Arrays.stream(String.join(",", numbers)
+                        .split(","))
+                        .map(Integer::parseInt)
+                        .filter(n -> n % 2 == 0)
+                        .mapToInt(n -> n)
+                        .min();
+        return i.getAsInt();
```
MaliukDaria, line 31:
```suggestion
       return numbers.stream()
                        .flatMap(n -> Arrays.stream(n.split(",")))
                        .map(Integer::valueOf)
                        .filter(n -> n % 2 == 0)
                        .mapToInt(n -> n)
                        .min()
                        .orElseThrow((() -> new RuntimeException("Can't get min value from list " + numbers)));
```

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #335, opened 2022-06-02
implement task

gdpbalt:
- add implementation of Predicate interface in class CandidateValidator;
- add implementation of tasks in class StreamPractice.

### DIFF
--- a/src/main/java/practice/CandidateValidator.java
+++ b/src/main/java/practice/CandidateValidator.java
@@ -1,5 +1,28 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int AGE_MINIMUM_VALID = 35;
+    private static final String NATIONALITY_VALID = "ukrainian";
+    private static final String PERIODS_IN_COUNTRY_SPLIT = "-";
+    private static final int INDEX_BEGIN_IN_COUNTRY = 0;
+    private static final int INDEX_END_IN_COUNTRY = 1;
+
+    @Override
+    public boolean test(Candidate candidate) {
+        if (candidate.getAge() < AGE_MINIMUM_VALID) {
+            return false;
+        }
+        if (!candidate.isAllowedToVote()) {
+            return false;
+        }
+        if (!candidate.getNationality().equalsIgnoreCase(NATIONALITY_VALID)) {
+            return false;
+        }
+        String[] partsPeriodsInUkr = candidate.getPeriodsInUkr().split(PERIODS_IN_COUNTRY_SPLIT);
+        return (Integer.parseInt(partsPeriodsInUkr[INDEX_END_IN_COUNTRY])
+                - Integer.parseInt(partsPeriodsInUkr[INDEX_BEGIN_IN_COUNTRY])) >= 10;
+    }
 }
--- a/src/main/java/practice/StreamPractice.java
+++ b/src/main/java/practice/StreamPractice.java
@@ -1,11 +1,19 @@
 package practice;
 
-import java.util.Collections;
+import java.util.Comparator;
 import java.util.List;
+import java.util.Optional;
+import java.util.function.Predicate;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import model.Candidate;
+import model.Cat;
 import model.Person;
 
 public class StreamPractice {
+    private static final String SPLIT_PATTERN = ",";
+
     /**
      * Given list of strings where each element contains 1+ numbers:
      * input = {"5,30,100", "0,22,7", ...}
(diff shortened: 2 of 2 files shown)


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,28 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int AGE_MINIMUM_VALID = 35;
+    private static final String NATIONALITY_VALID = "ukrainian";
```
viktoriia-haraniuk, line 8:
```suggestion
    private static final String NATIONALITY_VALID = "Ukrainian";
```

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,28 @@
...
+        }
+        if (!candidate.isAllowedToVote()) {
+            return false;
+        }
+        if (!candidate.getNationality().equalsIgnoreCase(NATIONALITY_VALID)) {
+            return false;
+        }
+        String[] partsPeriodsInUkr = candidate.getPeriodsInUkr().split(PERIODS_IN_COUNTRY_SPLIT);
+        return (Integer.parseInt(partsPeriodsInUkr[INDEX_END_IN_COUNTRY])
+                - Integer.parseInt(partsPeriodsInUkr[INDEX_BEGIN_IN_COUNTRY])) >= 10;
```
viktoriia-haraniuk, line 26: 10 should be constant

viktoriia-haraniuk, line 26: You can simplify all this construction using boolean operators.

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +22,13 @@ public class StreamPractice {
...
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        Optional<Integer> min = numbers.stream()
+                .flatMap(Pattern.compile(SPLIT_PATTERN)::splitAsStream)
+                .map(Integer::parseInt)
+                .filter(n -> n % 2 == 0)
+                .min(Comparator.naturalOrder());
+        return min.orElseThrow(() -> new RuntimeException(
+                String.format("Can't get min value from list: %s", numbers)));
```
viktoriia-haraniuk, line 31:
```suggestion
        return numbers.stream()
                .flatMap(Pattern.compile(SPLIT_PATTERN)::splitAsStream)
                .map(Integer::parseInt)
                .filter(n -> n % 2 == 0)
                .min(Comparator.naturalOrder())
                .orElseThrow(() -> new RuntimeException(
                String.format("Can't get min value from list: %s", numbers)));
```

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +73,18 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
...
-        return Collections.emptyList();
+        Predicate<Person> validator = p -> {
+            if (p.getAge() < fromAge) {
+                return false;
+            }
+            if (p.getSex() == Person.Sex.MAN && p.getAge() <= maleToAge) {
+                return true;
+            }
+            return p.getSex() == Person.Sex.WOMAN && p.getAge() <= femaleToAge;
+        };
```
viktoriia-haraniuk, line 84: You can simplify this construction with boolean operations.

viktoriia-haraniuk (changes requested):
Almost perfect 😸

src/main/java/practice/CandidateValidator.java
gdpbalt: > 

I use equalsIgnoreCase in this case.
As result the first letter of the nationality can be Upper or Lower case

viktoriia-haraniuk: image
As I understood from task, nationality always look like that, but it's up to you how to name this field)

viktoriia-haraniuk (approved):
Great 😄

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +36,12 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .mapToObj(i -> i % 2 > 0 ? numbers.get(i) - 1 : numbers.get(i))
+                .mapToInt(i -> i)
```
KostyaNikiforov, line 41:
```suggestion
                .map(i -> i % 2 > 0 ? numbers.get(i) - 1 : numbers.get(i))
```

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):

PR #336, opened 2022-06-06
Homework stream practice

Cosmic-M:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,73 @@
...
+        List<String> resultList = numbers.stream()
+                .map(i -> i + ",")
+                .collect(Collectors.toList());
+        return Arrays.stream(String.join("", resultList)
+                        .split(","))
+                .map(Integer::parseInt)
+                .filter(i -> i % 2 == 0)
+                .min(Comparator.naturalOrder())
+                .orElseThrow(() -> new RuntimeException(
+                        "Can't get min value from list: " + numbers));
```
Sm0kk, line 28: Let's think how you can do this in one stream() ? 
For throwing exception you can use .orElseThrow() method.

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,73 @@
...
-     * Example: select people of working age
-     * (from 18 y.o. and to 60 y.o. for men and to 55 y.o. for women inclusively).
-     */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(i -> i.getAge() >= fromAge)
+                .filter(i -> i.getSex() == Person.Sex.MAN
+                        ? i.getAge() <= maleToAge : i.getAge() <= femaleToAge)
```
Sm0kk, line 51: Here you need to use some predicate.

Sm0kk (changes requested):

Sm0kk (approved):

liliia-ponomarenko (changes requested):
Good solution! Let’s improve it 😉

PR #337, opened 2022-06-10
Stream methods implemented

antonEEmus:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,65 @@
...
-     * Given list of strings where each element contains 1+ numbers:
-     * input = {"5,30,100", "0,22,7", ...}
-     * return min integer value. One more thing - we're interested in even numbers.
-     * If there is no needed data throw RuntimeException with message
-     * "Can't get min value from list: < Here is our input 'numbers' >"
-     */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(n -> Arrays.stream(n.length() > 1 ? n.split(",") : new String[] {n}))
```
zim10093, line 15: Even if it has an empty list it will return  stream;
And do not hardcode, use constants for magic value.
```suggestion
                .flatMap(n -> Arrays.stream(n.split(",")))
```

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,65 @@
...
-     * new Person(«Helen», 42, Sex.WOMAN))`,
-     * select from the List only men whose age is from `fromAge` to `toAge` inclusively.
-     * <p>
-     * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
-     */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(n -> n.getAge() >= fromAge && n.getAge() <= toAge
+                        && n.getSex().equals(Person.Sex.MAN))
```
zim10093, line 38: You can compare Sex by "==" because enam is null safe
You did it a few times

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,65 @@
...
-     * <p>
-     * Example: select people of working age
-     * (from 18 y.o. and to 60 y.o. for men and to 55 y.o. for women inclusively).
-     */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(n -> n.getAge() >= fromAge && n.getAge()
+                        <= (n.getSex().equals(Person.Sex.MAN) ? maleToAge : femaleToAge))
```
zim10093, line 46: It is would better if you make a separate Predicate and use it in the filter method

zim10093 (changes requested):

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,23 @@
...
+
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= MINIMUM_AGE
+                && candidate.getNationality().equals(NATIONALITY)
+                && candidate.isAllowedToVote()
+                && Integer.parseInt(candidate.getPeriodsInUkr()
+                .split(PERIOD_SEPARATOR)[END_OF_PERIOD_INDEX])
+                - Integer.parseInt(candidate.getPeriodsInUkr()
+                        .split(PERIOD_SEPARATOR)[START_OF_PERIOD_INDEX]) >= MINIMUM_PERIOD;
+    }
```
zim10093, line 22: Better to check in a different method.
So I know that you prefer writing code in one method of simply checking, but here you have a huge part of checking period of live in Ukraine in one method.
I don`t know, but I wrote this comment in the last review and it doesn't save.

zim10093 (changes requested):

zim10093 (approved):
I think checkPeriods may split into a few lines of code, but I don`t know if you must do it

KostyaNikiforov (approved):
LGTM:fire:

PR #339, opened 2022-06-10
Jv stream practice

Zinchuk-Yevhenii-Mate:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,21 @@
...
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MIN_AGE_FOR_RUNNING = 35;
+    private static final int PERIOD_LIVING_IN_UKRAINE = 10;
+    private static final String NATIONALITY_FOR_RUNNING = "Ukrainian";
+
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.getClass().equals(Candidate.class)
```
ChumakAndrey, line 13: we don't need to check class here. If class will be wrong we'll get compile error

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,21 @@
...
+                && candidate.getAge() >= MIN_AGE_FOR_RUNNING
+                && candidate.getNationality().equals(NATIONALITY_FOR_RUNNING)
+                && candidate.isAllowedToVote()
+                && (Integer.parseInt(candidate.getPeriodsInUkr().substring(5))
+                - Integer.parseInt(candidate.getPeriodsInUkr().substring(0, 4)))
```
ChumakAndrey, line 18: Magic numbers

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,67 @@
...
-     * return the average of all odd numbers from the list or throw NoSuchElementException.
-     * But before that subtract 1 from each element on an odd position (having the odd index).
-     */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(index -> index % 2 != 0 ? numbers.get(index) - 1 : numbers.get(index))
+                .filter(value -> value % 2 != 0)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
ChumakAndrey, line 30: You should use `getAsDouble()`. It return double or Exception

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,67 @@
...
-     * for example, `Arrays.asList( new Person(«Victor», 16, Sex.MAN),
-     * new Person(«Helen», 42, Sex.WOMAN))`,
-     * select from the List only men whose age is from `fromAge` to `toAge` inclusively.
-     * <p>
-     * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
-     */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(people -> people.getSex().equals(Person.Sex.MAN)
```
ChumakAndrey, line 35: It better to use `==` to compare ENUM. It works faster

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,67 @@
...
-     * (from 18 y.o. and to 60 y.o. for men and to 55 y.o. for women inclusively).
-     */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(people -> (people.getSex().equals(Person.Sex.MAN)
+                        && people.getAge() >= fromAge && people.getAge() <= maleToAge)
+                        || (people.getSex().equals(Person.Sex.WOMAN)
+                        && people.getAge() >= fromAge && people.getAge() <= femaleToAge))
```
ChumakAndrey, line 46: You can use Predicate to create filtering parameters.

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,67 @@
...
 
-    /**
-     * Given a List of `Person` instances (having `name`, `age`, `sex` and `cats` fields,
-     * and each `Cat` having a `name` and `age`),
-     * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
-     */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter((people -> people.getSex()
```
ChumakAndrey, line 52: == with ENUM

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,67 @@
...
+                        .equals(Person.Sex.WOMAN) && people.getAge() >= femaleAge
+                        && !people.getCats().isEmpty()))
+                .flatMap(person -> person.getCats().stream().map(Cat::getName))
```
ChumakAndrey, line 55: Call each new method from a new line in the stream.

ChumakAndrey (changes requested):

ChumakAndrey (approved):

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,22 @@
...
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MIN_AGE_FOR_RUNNING = 35;
+    private static final int MIN_PERIOD_LIVING_IN_UKRAINE = 10;
+    private static final String NATIONALITY_FOR_RUNNING = "Ukrainian";
+    private static final String SEPARATOR = "-";
+
+    @Override
+    public boolean test(Candidate candidate) {
+        int fromAge = Integer.parseInt(candidate.getPeriodsInUkr().split(SEPARATOR)[0]);
+        int toAge = Integer.parseInt(candidate.getPeriodsInUkr().split(SEPARATOR)[1]);
```
KostyaNikiforov, line 15: Here you are duplicating the code by calling the split method twice instead reusing its result

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,70 @@
...
-     * and live in Ukraine for 10 years. For the last requirement use field periodsInUkr,
-     * which has following view: "2002-2015"
-     * We want to reuse our validation in future, so let's write our own impl of Predicate
-     * parametrized with Candidate in CandidateValidator.
-     */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        CandidateValidator candidateValidator = new CandidateValidator();
+        return candidates.stream()
+                .filter(candidateValidator::test)
```
KostyaNikiforov, line 65:
```suggestion
                .filter(new CandidateValidator())
```

KostyaNikiforov (changes requested):

src/main/java/practice/CandidateValidator.java
```diff
@@ -11,8 +11,9 @@ public class CandidateValidator implements Predicate<Candidate> {
 
     @Override
     public boolean test(Candidate candidate) {
-        int fromAge = Integer.parseInt(candidate.getPeriodsInUkr().split(SEPARATOR)[0]);
-        int toAge = Integer.parseInt(candidate.getPeriodsInUkr().split(SEPARATOR)[1]);
+        String[] periodInUkraineArray = candidate.getPeriodsInUkr().split(SEPARATOR);
+        int fromAge = Integer.parseInt(periodInUkraineArray[0]);
+        int toAge = Integer.parseInt(periodInUkraineArray[1]);
```
KostyaNikiforov, line 16: It is better to make those values constants too

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):

PR #343, opened 2022-06-14
Hw stream practice

nazarbaklanov:


Pavlo99-gif (changes requested):
need small correct

Pavlo99-gif (approved):

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,66 @@
...
-     * have nationality - 'Ukrainian'
-     * and live in Ukraine for 10 years. For the last requirement use field periodsInUkr,
-     * which has following view: "2002-2015"
-     * We want to reuse our validation in future, so let's write our own impl of Predicate
-     * parametrized with Candidate in CandidateValidator.
-     */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(p -> new CandidateValidator().test(p))
```
KostyaNikiforov, line 61:
```suggestion
                .filter(new CandidateValidator())
```

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,66 @@
...
-     * (from 18 y.o. and to 60 y.o. for men and to 55 y.o. for women inclusively).
-     */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getAge() >= fromAge
+                        && ((p.getSex().equals(Person.Sex.MAN)
+                        && p.getAge() <= maleToAge)
+                        || (p.getSex().equals(Person.Sex.WOMAN) && p.getAge() <= femaleToAge)))
```
KostyaNikiforov, line 47: It would be cool to create a predicate for huge conditions

KostyaNikiforov (changes requested):

src/main/java/practice/StreamPractice.java
nazarbaklanov: It makes the code worse to read. Doesn't it?

KostyaNikiforov: In this case, it's hard to disagree. But sometimes it's useful because you can add extra context to the condition by giving it a name

KostyaNikiforov (approved):

src/main/java/practice/StreamPractice.java
nazarbaklanov: Ok, I understand.

PR #344, opened 2022-06-14
All methods were implements, all tests run clear.

MKaramavrov:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,27 @@
...
+        int yearFrom = Integer.parseInt(temp[GET_DATE_FROM]);
+        int yearTo = Integer.parseInt(temp[GET_DATE_TO]);
+        return (yearTo - yearFrom) >= ELIGIBLE_PERIOD;
+    }
+
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.isAllowedToVote()
+                && candidate.getAge() >= ELIGIBLE_AGE
+                && getYearsInUkraine(candidate)
```
Cosmic-M, line 24: I propose you delete method getYearsInUkraine, a couple constants - GET_DATE_FROM, GET_DATE_TO (in my opinion it's not so magic numbers). And do all logic in one method. It will be shorter:
        String[] temp = candidate.getPeriodsInUkr().split("-");
        return candidate.isAllowedToVote()
                && candidate.getAge() >= ELIGIBLE_AGE
                && (Integer.parseInt(temp[1]) - Integer.parseInt(temp[0]) > ELIGIBLE_PERIOD)
                && candidate.getNationality().equals(ELIGIBLE_NATIONALITY);
    }

KostyaNikiforov (approved):

PR #350, opened 2022-06-18
stream practise is ready

KarinaTanasiichuk:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,21 @@
...
+
+    @Override
+    public boolean test(Candidate candidate) {
+        int firstDate = Integer.parseInt(candidate.getPeriodsInUkr()
+                .substring(0, candidate.getPeriodsInUkr().indexOf("-")));
+        int lastDate = Integer.parseInt(candidate.getPeriodsInUkr()
+                .substring(candidate.getPeriodsInUkr().indexOf("-") + 1));
+        return (candidate.getAge() >= MIN_AGE) && candidate.isAllowedToVote()
+                && (candidate.getNationality().equals(NATIONALITY))
+                && ((lastDate - firstDate) > MIN_YEARS);
```
olyaklipa, line 19: I would put here ((lastDate - firstDate) >= MIN_YEARS); Because README file says: "live in urk for 10 years", so I think 10 should be included

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,21 @@
...
+    private static final int MIN_AGE = 35;
+    private static final int MIN_YEARS = 10;
+    private static final String NATIONALITY = "Ukrainian";
+
+    @Override
+    public boolean test(Candidate candidate) {
+        int firstDate = Integer.parseInt(candidate.getPeriodsInUkr()
+                .substring(0, candidate.getPeriodsInUkr().indexOf("-")));
+        int lastDate = Integer.parseInt(candidate.getPeriodsInUkr()
+                .substring(candidate.getPeriodsInUkr().indexOf("-") + 1));
```
olyaklipa, line 16: Alternatively, you can move this logic into a separate method. But it's up to you.

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +36,11 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(i -> (i % 2 != 0) ? numbers.get(i) - 1 : numbers.get(i))
+                .filter(k -> k % 2 != 0)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
olyaklipa, line 43: .getAsDouble() - if you use this one you don't need to throw that exception explicitly

olyaklipa (changes requested):

olyaklipa (approved):
Great job!

liliia-ponomarenko (approved):
LGTM 🔥

PR #351, opened 2022-06-19
Tasks were done

olyaklipa:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,19 @@
...
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= 35
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals("Ukrainian")
+                && numberOfYearsInUkr(candidate.getPeriodsInUkr()) >= 10;
```
KarinaTanasiichuk, line 12: I think, you could use constants with these magic numbers;)

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +24,14 @@ public class StreamPractice {
...
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        OptionalInt value = numbers.stream()
+                .map(w -> w.split(","))
+                .flatMap(Arrays::stream)
+                .mapToInt(Integer::parseInt)
+                .filter(n -> n % 2 == 0)
+                .min();
+        return value.orElseThrow(() -> new RuntimeException("Can't get min value from list: "
```
KarinaTanasiichuk, line 33: We can return result without creating a new object))

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +40,11 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        OptionalDouble value = IntStream.range(0, numbers.size())
+                .map(i -> i % 2 == 0 ? numbers.get(i) : numbers.get(i) - 1)
+                .filter(i -> i % 2 == 1)
+                .average();
+        return value.getAsDouble();
```
KarinaTanasiichuk, line 47: The same thing;)

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +56,11 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        List<Person> selected = peopleList.stream()
+                .filter(p -> p.getSex() == MAN
```
KarinaTanasiichuk, line 60: You can create here your custom Predicate implementation, and than use it) 
If you want, of course)

KarinaTanasiichuk (approved):

src/main/java/practice/StreamPractice.java
liliia-ponomarenko: agree

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +95,14 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        List<String> selected = peopleList.stream()
+                .filter(p -> p.getSex() == WOMAN && p.getAge() >= femaleAge)
+                .map(Person::getCats)
+                .flatMap(Collection::stream)
+                .map(Cat::getName)
+                .distinct()
```
liliia-ponomarenko, line 103:
```suggestion
```

liliia-ponomarenko (approved):
good job 👍

PR #352, opened 2022-06-19
implemented streams

DimaPodakin:


src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +35,11 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(i -> i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i))
+                .filter(number -> number % 2 != 0)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
YaroslavLa, line 42: If you need to throw an exception when no result is found. You can use get() or getAsDouble()

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +82,13 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getAge() >= femaleAge
+                        && person.getSex().equals(Person.Sex.WOMAN))
```
YaroslavLa, line 87: "==" is a better way to compare Enum values

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,14 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(s -> s.split(","))
```
YaroslavLa, line 23: Let`s make "," constant

YaroslavLa (changes requested):

YaroslavLa (approved):
LGTM!

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +70,11 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getAge() >= fromAge
```
liliia-ponomarenko, line 74: Let’s use Predicate 
image

liliia-ponomarenko (changes requested):
Good solution! Let’s improve it 😉

liliia-ponomarenko (approved):
good job 👍

PR #367, opened 2022-07-11
Homework Stream Practice

Mykola-Kutsyi:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,21 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int PERIOD_START_ID = 0;
+    private static final int PERIOD_END_ID = 1;
```
```suggestion
    private static final int PERIOD_START_IINDEX = 0;
    private static final int PERIOD_END_INDEX = 1;
```

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,21 @@
...
+    private static final int MIN_IN_UKRAINE = 10;
```
MaliukDaria, line 9:
```suggestion
    private static final int MIN_YEARS_IN_UKRAINE = 10;
```

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,66 @@
...
-     * input = {"5,30,100", "0,22,7", ...}
-     * return min integer value. One more thing - we're interested in even numbers.
-     * If there is no needed data throw RuntimeException with message
-     * "Can't get min value from list: < Here is our input 'numbers' >"
-     */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(s -> s.split(","))
+                .flatMap(Arrays::stream)
```
MaliukDaria, line 16: could we combine map and flatMap into one flatMap?

MaliukDaria (approved):

PR #374, opened 2022-07-28
hw-stream-practice

RomanPlk:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,70 @@
...
-     * return min integer value. One more thing - we're interested in even numbers.
-     * If there is no needed data throw RuntimeException with message
-     * "Can't get min value from list: < Here is our input 'numbers' >"
-     */
+
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(s -> s.split(","))
+                .flatMap(Arrays::stream)
```
MaliukDaria, line 17: could we combine map and flatMap into one flatMap here?

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,70 @@
...
-     * But before that subtract 1 from each element on an odd position (having the odd index).
-     */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(i -> i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i))
+                .filter(i -> i % 2 != 0)
+                .mapToDouble(Double::valueOf)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
MaliukDaria, line 31: what  `orElseThrow` method throws by default?

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,70 @@
...
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex().equals(Person.Sex.MAN)
+                        && p.getAge() >= fromAge
+                        && p.getAge() <= maleToAge
+                        || p.getSex().equals(Person.Sex.WOMAN)
+                        && p.getAge() >= fromAge
+                        && p.getAge() <= femaleToAge)
```
MaliukDaria, line 50: could we provide `p.getAge() >= fromAge` check only ones here?

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #376, opened 2022-07-29
J c 20 5

TurchynSergii:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,14 @@ public class StreamPractice {
...
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        Optional<Integer> min = numbers.stream()
+                .map(s -> s.split(","))
+                .flatMap(Arrays::stream)
+                .map(Integer::parseInt)
+                .filter(i -> i % 2 == 0)
+                .min(Integer::compareTo);
+        return min.orElseThrow(() ->
+                new RuntimeException("Can't get min value from list: " + numbers));
```
```suggestion
        return numbers.stream()
                .map(s -> s.split(","))
                .flatMap(Arrays::stream)
                .map(Integer::parseInt)
                .filter(i -> i % 2 == 0)
                .min(Integer::compareTo);
                .orElseThrow(() ->
                new RuntimeException("Can't get min value from list: " + numbers));
```

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,14 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        Optional<Integer> min = numbers.stream()
+                .map(s -> s.split(","))
+                .flatMap(Arrays::stream)
```
MaliukDaria, line 25: could we combine these map and flatMap into one flatMap?

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +90,14 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex() == Person.Sex.WOMAN
+                        && p.getAge() >= femaleAge)
+                .map(Person::getCats)
+                .flatMap(Collection::stream)
```
MaliukDaria, line 97: same, could we combine these map and flatMap into one flatMap?

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +36,12 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(i -> i % 2 == 1 ? numbers.get(i) - 1 : numbers.get(i))
+                .mapToDouble(i -> i)
+                .filter(i -> i % 2 == 1)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
MaliukDaria, line 44: What `orElseThrow` method throws by default?

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +72,16 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
...
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> {
+                    if (p.getSex() == Person.Sex.MAN) {
+                        return p.getAge() >= fromAge
+                                && p.getAge() <= maleToAge;
+                    }
+                    return p.getAge() >= fromAge
+                            && p.getAge() <= femaleToAge;
```
MaliukDaria, line 82: could we simplify this?

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #388, opened 2022-08-18
Hw solution is developed

amarkelovUA:


src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +33,18 @@ public int findMinEvenNumber(List<String> numbers) {
...
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return numbers.stream()
+                .map(number -> {
+                    if (isOdd) {
+                        number--;
+                    }
+                    isOdd = !isOdd;
+                    return number;
+                })
```
Oleksandr-Tovstyi, line 43:
```suggestion
                .map(number -> {
                    if (isOdd) {
                        number--;
                    }
                    isOdd = !isOdd;
                    return number;
                })
```I used counter here, your solution looks better than mine.

Comment (Oleksandr-Tovstyi) [inline src/main/java/practice/StreamPractice.java:43]:
```suggestion
```
I used counter here, your solution looks better than mine.

Oleksandr-Tovstyi (approved):

src/main/java/practice/StreamPractice.java
```diff
@@ -1,11 +1,15 @@
...
-import java.util.Collections;
+import java.util.Arrays;
 import java.util.List;
+import java.util.stream.Collectors;
 import model.Candidate;
+import model.Cat;
 import model.Person;
 
 public class StreamPractice {
+    private boolean isOdd = false;
```
Sofasmile, line 11: let's move this variable to the method

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +33,18 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return numbers.stream()
```
Sofasmile, line 36: or you can do this task using `IntStream.range(...)`

Sofasmile (changes requested):

Sofasmile (approved):

PR #401, opened 2022-08-23
Implemented methods in StreamPractice. Implemented by CandidateValida…

justpopova:
…tor Predicate and implemented the logic into test method.

### DIFF
--- a/src/main/java/practice/CandidateValidator.java
+++ b/src/main/java/practice/CandidateValidator.java
@@ -1,5 +1,25 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MIN_AGE = 35;
+    private static final String NATIONALITY = "Ukrainian";
+    private static final int YEARS_SPENT_IN_UA = 10;
+    private static final int YEAR_FROM = 0;
+    private static final int YEAR_TO = 1;
+
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= MIN_AGE && candidate.isAllowedToVote()
+                && candidate.getNationality().equals(NATIONALITY)
+                && totalYearsSpentInUA(candidate.getPeriodsInUkr()) >= YEARS_SPENT_IN_UA;
+    }
+
+    private int totalYearsSpentInUA(String periodsInUA) {
+        String[] splittedYears = periodsInUA.split("-");
+        return Integer.parseInt(splittedYears[YEAR_TO])
+                - Integer.parseInt(splittedYears[YEAR_FROM]);
+    }
 }
--- a/src/main/java/practice/StreamPractice.java
+++ b/src/main/java/practice/StreamPractice.java
@@ -1,80 +1,68 @@
 package practice;
 
-import java.util.Collections;
+import java.util.Arrays;
+import java.util.Collection;
 import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import model.Candidate;
+import model.Cat;
 import model.Person;
 
 public class StreamPractice {
-    /**
-     * Given list of strings where each element contains 1+ numbers:
-     * input = {"5,30,100", "0,22,7", ...}
-     * return min integer value. One more thing - we're interested in even numbers.
-     * If there is no needed data throw RuntimeException with message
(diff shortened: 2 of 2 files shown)


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,25 @@
...
+
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= MIN_AGE && candidate.isAllowedToVote()
+                && candidate.getNationality().equals(NATIONALITY)
+                && totalYearsSpentInUA(candidate.getPeriodsInUkr()) >= YEARS_SPENT_IN_UA;
+    }
+
+    private int totalYearsSpentInUA(String periodsInUA) {
+        String[] splittedYears = periodsInUA.split("-");
```
AshaJaneBrown, line 21: "-" should be constant as well, e.g. "separator" or smth like that

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,68 @@
...
-     * new Person(«Helen», 42, Sex.WOMAN))`,
-     * select from the List only men whose age is from `fromAge` to `toAge` inclusively.
-     * <p>
-     * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
-     */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(m -> m.getAge() >= fromAge && m.getAge() <= toAge)
+                .filter(s -> s.getSex().equals(Person.Sex.MAN))
```
AshaJaneBrown, line 38: commonly, we use 1 filter instead of 2 if filtering logic is the same - take a person and test it

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,68 @@
...
-     * or to `femaleToAge` (for women) inclusively.
-     * <p>
-     * Example: select people of working age
-     * (from 18 y.o. and to 60 y.o. for men and to 55 y.o. for women inclusively).
-     */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> (p.getAge() >= fromAge && p.getAge() <= femaleToAge
```
AshaJaneBrown, line 45: due to mentors' comments, we need to do this check p.getAge() >= fromAge only once, for every person in the list, and then conbine this check with other checks via &&

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,68 @@
...
-     * Given a List of `Person` instances (having `name`, `age`, `sex` and `cats` fields,
-     * and each `Cat` having a `name` and `age`),
-     * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
-     */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(owner -> owner.getAge() >= femaleAge && owner.getSex() == Person.Sex.WOMAN)
+                .map(Person::getCats)
+                .flatMap(Collection::stream)
```
AshaJaneBrown, line 56: due to mentors comments, we could merge this two operations (.map and .flatmap) in one - .flatmap

AshaJaneBrown (changes requested):
Great job!

AshaJaneBrown (approved):

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,66 @@
...
 
-    /**
-     * Given a List of Integer numbers,
-     * return the average of all odd numbers from the list or throw NoSuchElementException.
-     * But before that subtract 1 from each element on an odd position (having the odd index).
-     */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .mapToObj(i -> (i % 2 == 0) ? numbers.get(i) : numbers.get(i) - 1)
```
Noct2000, line 27: if you replace `mapToObj ` with `map` you can remove `.mapToDouble(n -> (double) n)`
```suggestion
                .mapToObj(i -> i % 2 == 0 ? numbers.get(i) : numbers.get(i) - 1)
```

Noct2000 (changes requested):

Noct2000 (approved):

PR #403, opened 2022-08-23
added solution

Nesterenko-Andrey:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,22 @@
...
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int ELIGIBLE_AGE = 35;
+    private static final int INDEX_OF_AGE_FROM = 0;
+    private static final int INDEX_OF_AGE_TO = 1;
```
GolDenik, line 9: I think it to be better if you rename it like :
```suggestion
    private static final int INDEX_OF_YEAR_FROM = 0;
    private static final int INDEX_OF_YEAR_TO = 1;
```

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,22 @@
...
+    private static final int ELIGIBLE_PERIOD = 10;
+    private static final String ELIGIBLE_NATIONALITY = "Ukrainian";
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] temp = candidate.getPeriodsInUkr().split("-");
```
GolDenik, line 15: I am not sure , but maybe you should add some constant for this like DATA_DELIMITER or something else
```suggestion
        String[] temp = candidate.getPeriodsInUkr().split(SOME_CONSTANT);
```

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +69,12 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> ((p.getSex() == Person.Sex.MAN
+                        && p.getAge() >= fromAge && p.getAge() <= maleToAge))
+                        || (p.getSex() == Person.Sex.WOMAN
+                        && p.getAge() >= fromAge && p.getAge() <= femaleToAge))
```
GolDenik, line 76: You used "p.getAge() >= fromAge" 2 times , but you can use it only once

GolDenik (changes requested):

src/main/java/practice/StreamPractice.java
Nesterenko-Andrey: agree but I think it easier for reading in such way)

GolDenik (approved):

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,14 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(s -> s.split(","))
+                .flatMap(Arrays::stream)
```
Noct2000, line 24: you can use `s.split(",")` in the `.flatMap`

src/main/java/practice/StreamPractice.java
Noct2000, line 76: you can check `fromAge` only once

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +83,12 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex() == Person.Sex.WOMAN && p.getAge() >= femaleAge)
+                .map(Person::getCats)
+                .flatMap(Collection::stream)
```
Noct2000, line 89: you can combine operations in the `.flatMap`

Noct2000 (changes requested):

Noct2000 (approved):

PR #404, opened 2022-08-23
all tests passed, checkstyle ok

max-bolotin:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,15 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator<T> implements Predicate<T> {
```
shevalye, line 6: this can be simplified  to CandidateValidator \<Candidate \>  so that there is no need for casting below.

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,15 @@
...
+    @Override
+    public boolean test(T t) {
+        Candidate candidate = (Candidate) t;
+        int yearsInUkraine = Integer.parseInt(candidate.getPeriodsInUkr().substring(5, 9))
+                - Integer.parseInt(candidate.getPeriodsInUkr().substring(0, 4));
+        return candidate.getAge() >= 35 && candidate.isAllowedToVote()
+                && "Ukrainian".equals(candidate.getNationality()) && yearsInUkraine >= 10;
```
shevalye, line 13: we should use constants here not literals, magic numbers

shevalye (changes requested):

shevalye (approved):

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,19 @@
...
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final String NATIONALITY = "Ukrainian";
+    private static final int MIN_AGE = 35;
+    private static final int MIN_YEARS_IN_UKRAINE = 10;
+
+    @Override
+    public boolean test(Candidate candidate) {
+        int yearsInUkraine = Integer.parseInt(candidate.getPeriodsInUkr().substring(5, 9))
+                - Integer.parseInt(candidate.getPeriodsInUkr().substring(0, 4));
```
Noct2000, line 14: you can use `split("-")` and parse elements to int

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,16 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(s -> s.split(","))
+                .map(Arrays::asList)
+                .flatMap(Collection::stream)
```
Noct2000, line 25: you can use `.split(",")` in the .flatMap, and you can use `Arrays.stream`

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,16 @@ public class StreamPractice {
...
+                .mapToInt(Integer::parseInt)
+                .filter(n -> n % 2 == 0)
+                .sorted()
+                .findFirst()
```
Noct2000, line 29: you can use `min()` from `IntStream`

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +37,12 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+
+        return IntStream.range(0, numbers.size())
+                .map(index -> (index % 2 == 0) ? numbers.get(index) : numbers.get(index) - 1)
```
Noct2000, line 42:
```suggestion
                .map(index -> index % 2 == 0 ? numbers.get(index) : numbers.get(index) - 1)
```

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +88,15 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
...
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList
+                .stream()
+                .filter(p -> p.getAge() >= femaleAge && p.getSex().equals(Person.Sex.WOMAN))
+                .map(p -> p.getCats()
+                        .stream()
+                        .map(Cat::getName)
+                        .collect(Collectors.toList()))
+                .flatMap(List::stream)
```
Noct2000, line 98: your `.map` is too complex, lets use `p -> p.getCatList().stream()` in the `.flatMap` and map cats to names later

Noct2000 (changes requested):

src/main/java/practice/CandidateValidator.java
max-bolotin: Yes I can, and I'll do it, I just don't understand why is it better than substring? My solution is in one line while solution via split needs two lines

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,19 @@
...
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final String NATIONALITY = "Ukrainian";
+    private static final int MIN_AGE = 35;
+    private static final int MIN_YEARS_IN_UKRAINE = 10;
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] years = candidate.getPeriodsInUkr().split("-");
+        int yearsInUkraine = Integer.parseInt(years[1]) - Integer.parseInt(years[0]);
+        return candidate.getAge() >= MIN_AGE && candidate.isAllowedToVote()
```
Noct2000, line 15: replace all indexes and `"-"` with constants

Noct2000 (changes requested):

Noct2000 (approved):
Good job) 🔥

PR #416, opened 2022-08-25
Was completed stream practice

valentyn-slutskyi:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,14 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(s -> s.split(","))
+                .flatMap(Arrays::stream)
```
JustElvis, line 25: flatMap(s -> Stream.of(s.split(",")))

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,14 @@ public class StreamPractice {
...
+                .mapToInt(Integer::parseInt)
+                .filter(n -> n % 2 == 0)
+                .min()
```
JustElvis, line 28: .min(Integer::compareTo)

JustElvis (changes requested):

src/main/java/practice/StreamPractice.java
valentyn-slutskyi: Thanks, it really works!

src/main/java/practice/StreamPractice.java
valentyn-slutskyi: Sorry, but it doesn't work in my case. Also, it does not look easier, so I just left as it was.

JustElvis (approved):

Sofasmile (approved):

PR #434, opened 2022-08-26
Hw java8 2 1

LubusUzh:


src/main/java/practice/WorkablePeople.java
```diff
@@ -0,0 +1,25 @@
+package practice;
+
+import java.util.function.Predicate;
+import model.Person;
+
+public class WorkablePeople implements Predicate<Person> {
+    private int fromAge;
```
JK-devv, line 7: my opinion that you can delete all fields and constructor 
and left just one method - test (Person person)
cuz person - already have needed fields for your predicate

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,61 @@
...
-     * or to `femaleToAge` (for women) inclusively.
-     * <p>
-     * Example: select people of working age
-     * (from 18 y.o. and to 60 y.o. for men and to 55 y.o. for women inclusively).
-     */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(new WorkablePeople(fromAge, femaleToAge, maleToAge))
```
JK-devv, line 42: and if you agree with my comment in your class WorckablePeople, than you can write in filter just: new WorcablePeople
:)

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,61 @@
...
 
-    /**
-     * Given a List of `Person` instances (having `name`, `age`, `sex` and `cats` fields,
-     * and each `Cat` having a `name` and `age`),
-     * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
-     */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(new CatOwnersFilter(femaleAge))
```
JK-devv, line 48: the same opinion

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,61 @@
...
-     * and live in Ukraine for 10 years. For the last requirement use field periodsInUkr,
-     * which has following view: "2002-2015"
-     * We want to reuse our validation in future, so let's write our own impl of Predicate
-     * parametrized with Candidate in CandidateValidator.
-     */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(new CandidateValidator())
+                .map(c -> c.getName())
```
JK-devv, line 57: .map(c -> c.getName()) - you can replace it with method reference 
  Candidate :: getName

JK-devv (changes requested):

JK-devv (approved):

src/main/java/practice/WorkablePeople.java
LubusUzh: I would like to do this, but unfortunately, Predicate works only with one type of data.

src/main/java/practice/StreamPractice.java
LubusUzh: I would like to do this, but unfortunately, Predicate works only with one type of data.

src/main/java/practice/StreamPractice.java
LubusUzh: I would like to do this, but unfortunately, Predicate works only with one type of data.

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,61 @@
...
-     * Given a List of `Person` instances (having `name`, `age`, `sex` and `cats` fields,
-     * and each `Cat` having a `name` and `age`),
-     * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
-     */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(new CatOwnersFilter(femaleAge))
+                .flatMap(p -> p.getCats().stream())
+                .map(c -> c.getName())
```
Noct2000, line 50: you can replace it with method reference

Noct2000 (approved):
it would be better to use lambda instead of CatOwnersFilter and WorkablePeople

PR #438, opened 2022-08-26
Solution

ghost1975:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,21 @@
...
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int OLDER_THAN = 35;
+    private static final int LIVE_IN_UKRAINE = 10;
+    private static final String NATIONALITY = "Ukrainian";
+    private static final String SPLITERATOR = "-";
```
aaananas, line 10:
```suggestion
    private static final int REQUIRED_MIN_AGE = 35;
    private static final int REQUIRED_YEARS_IN_UKRAINE = 10;
    private static final String REQUIRED_NATIONALITY = "Ukrainian";
    private static final String SPLITERATOR = "-";
```

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,21 @@
...
+
+    @Override
+    public boolean test(Candidate candidate) {
+        int period = Integer.parseInt(candidate.getPeriodsInUkr().split(SPLITERATOR)[1])
+                - Integer.parseInt(candidate.getPeriodsInUkr().split(SPLITERATOR)[0]);
```
aaananas, line 15:
```suggestion
        String[] splitted = candidate.getPeriodsInUkr().split(SPLITERATOR);
        int period = Integer.parseInt(splitted[PERIOD_TO_INDEX])
                - Integer.parseInt(splitted[PERIOD_FROM_INDEX]);
```

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +17,15 @@ public class StreamPractice {
...
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(s -> s.split(","))
+                .flatMap(Arrays::stream)
+                .mapToInt(Integer::valueOf)
+                .boxed()
+                .filter(i -> i % 2 == 0)
+                .min(Integer::compareTo)
+                .orElseThrow(() -> new RuntimeException("Can't get min value from list: "
```
```suggestion
                .flatMap(Arrays::stream)
                .mapToInt(Integer::valueOf)
                .filter(i -> i % 2 == 0)
                .min()
                .orElseThrow(() -> new RuntimeException("Can't get min value from list: "
```

aaananas (changes requested):

src/main/java/practice/CandidateValidator.java
aaananas: also change SPLITERATOR to SPLIT_COMMA

aaananas (approved):

PR #463, opened 2022-10-06
Jv stream GitHub practice

SoulARC:


SoulARC:
Some lags with git..

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,72 @@
...
-     */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(i -> i % 2 == 0
+                          ? numbers.get(i)
+                          : numbers.get(i) - 1)
+                .filter(i -> i % 2 != 0)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
KostyaNikiforov, line 31: 🔥

KostyaNikiforov (approved):
Good job):fire:

src/main/java/practice/StreamPractice.java
SoulARC: One of the few things I studied

PR #464, opened 2022-10-10
Stream practice solution v1

EvgenyTyshchenko:


src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +70,13 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getAge() >= fromAge)
+                .filter(p -> (p.getSex() == Person.Sex.MAN
```
liliia-ponomarenko, line 75: better use Predicate for long boolean expressions😉

liliia-ponomarenko (approved):
good job! 🔥

PR #469, opened 2022-10-20
StreamPractice implemented

hpavlo:


src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +37,14 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        AtomicInteger index = new AtomicInteger();
+        return numbers.stream()
+                .map(n -> n - (index.getAndIncrement() % 2 == 0 ? 0 : 1))
```
YuriiHryniv, line 42: I think that the AtomicInteger is not needed here. I used indexOf instead

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,22 @@
...
+    private static final int NEED_YEAR = 35;
+    private static final String NEED_NATIONALITY = "Ukrainian";
+    private static final int NEED_PERIOD_IN_UKRAINE = 10;
+    private static final String YEARS_SPLITTER = "-";
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] yearsInUkraine = candidate.getPeriodsInUkr().split(YEARS_SPLITTER);
+        int timeInUkraine = Integer.parseInt(yearsInUkraine[1])
+                - Integer.parseInt(yearsInUkraine[0]);
```
YuriiHryniv, line 16: I would move this logic to another private method

YuriiHryniv (approved):
It seems good to me, but I did something just differently

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,22 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int NEED_YEAR = 35;
```
liliia-ponomarenko, line 7:
```suggestion
    private static final int AVAILABLE_YEAR = 35;
```
NEED not really  appropriate use in this case

src/main/java/practice/CandidateValidator.java
liliia-ponomarenko, line 16: better make these indexes constants

liliia-ponomarenko (approved):
good job! 👍

PR #476, opened 2022-10-24
all tasks are done

MishaGetsko18:


src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +84,13 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(w -> w.getSex() == Person.Sex.WOMAN
+                        && w.getAge() >= femaleAge)
+                .flatMap(p -> p.getCats()
+                        .stream()
+                        .map(Cat::getName))
```
aaananas, line 92:
```suggestion
                .flatMap(p -> p.getCats().stream())
                .map(Cat::getName))
```

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +106,11 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        CandidateValidator validator = new CandidateValidator();
+        return candidates.stream()
+                .filter(validator)
```
aaananas, line 111:
```suggestion
        return candidates.stream()
                .filter(new CandidateValidator())
```
there is no need to create a separate variable

aaananas (changes requested):

zachar-chorny (approved):

PR #479, opened 2022-10-25
implemented StreamPractice and CandidateValidator

nazariipertsak:


Andew-Miroshnikov (approved):
Good job !

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,15 @@ public class StreamPractice {
...
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers
+                .stream()
+                .map(e -> e.split(","))
+                .flatMap(Arrays::stream)
+                .map(Integer::parseInt)
+                .filter(e -> e % 2 == 0)
+                .sorted()
+                .findFirst().orElseThrow(() ->
```
hronograf, line 29: it would be better to use `min` instead of `sorted` method

hronograf (approved):
Great! 🚀

PR #480, opened 2022-10-25
stream api practise

alex-anofriev:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,21 @@
...
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= 35 && candidate.isAllowedToVote()
+                && candidate.getNationality().equals("Ukrainian")
+                && getLifePeriod(candidate.getPeriodsInUkr()) >= 10;
```
GrishaSeniv, line 13: magic numbers...

GrishaSeniv (changes requested):

GrishaSeniv (approved):
good job)

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,26 @@
...
+        return candidate.getAge() >= MIN_AGE && candidate.isAllowedToVote()
+                && candidate.getNationality().equals(NATIONALITY)
+                && getLifePeriod(candidate.getPeriodsInUkr()) >= MIN_YEARS_IN_UA;
+    }
+
+    private int getLifePeriod(String years) {
+        return Arrays.stream(years.split(REGEX))
+                .mapToInt(Integer::parseInt)
+                .reduce((x,y) -> y - x)
+                .getAsInt();
```
liliia-ponomarenko, line 24: It's good that you had more practice with a stream, but in my opinion "everything  has a place".  in this case, it is less readable. but you can leave it if you want 😉

liliia-ponomarenko (changes requested):
Great solution! Let’s improve it 😉

liliia-ponomarenko (approved):
good job 😉🔥

PR #481, opened 2022-10-26
Stream Practice solution implement

anastasiavasylyk:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,23 @@
...
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals("Ukrainian")
+                && isValidPeriods(candidate);
+    }
+
+    private boolean isValidPeriods(Candidate candidate) {
+        String[] date = candidate.getPeriodsInUkr().split(SPLIT_FOR_DATE);
+        int yearFrom = Integer.parseInt(date[0]);
+        int yearTo = Integer.parseInt(date[1]);
+        return yearTo - yearFrom >= 10;
```
shmex-j, line 21: Magic numbers

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,23 @@
...
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final String SPLIT_FOR_DATE = "-";
+
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= 35
```
shmex-j, line 11: Also magic number

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +76,14 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
...
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> (person.getSex() == Person.Sex.MAN
+                        && person.getAge() >= fromAge
+                        && person.getAge() <= maleToAge)
+                        || (person.getSex() == Person.Sex.WOMAN
+                        && person.getAge() >= fromAge
+                        && person.getAge() <= femaleToAge))
```
shmex-j, line 85: (Optional) maybe you can simplify this condition

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +92,14 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
...
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        List<Cat> cats = peopleList.stream()
+                .filter(person -> person.getSex() == Person.Sex.WOMAN
+                        && person.getAge() >= femaleAge)
+                .flatMap(person -> person.getCats().stream())
+                .collect(Collectors.toList());
+        return cats.stream()
+                .map(Cat::getName)
+                .collect(Collectors.toList());
```
shmex-j, line 102: You can use single stream

src/main/java/practice/StreamPractice.java
shmex-j: also maybe you should [create your own predicate for this long condition](https://mate-academy.github.io/jv-program-common-mistakes/java-core/java-eight-part-two/stream-api#try-to-use-minimum-amount-of-stream-operations)

shmex-j (changes requested):
Some improvements

src/main/java/practice/StreamPractice.java
anastasiavasylyk: in this case I can`t use custom Predicate cause of fromAge, maleToAge and femaleToAge. Method test doesn`t accept more that one argument

shmex-j (approved):
Super

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +39,13 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(i -> i % 2 != 0
+                ? numbers.get(i) - 1
+                : numbers.get(i))
+                .filter(i -> i % 2 != 0)
```
hronograf, line 46: same about indent, `? numbers.get(i) - 1` should have indent that different from `.map`

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +76,13 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> (person.getSex() == Person.Sex.MAN
+                        && person.getAge() <= maleToAge
+                        || person.getSex() == Person.Sex.WOMAN
+                        && person.getAge() <= femaleToAge)
+                        && person.getAge() >= fromAge)
```
hronograf, line 84: let's create `Predicate` for this logic and then use it in `filter`

hronograf (changes requested):
Good!

src/main/java/practice/StreamPractice.java
anastasiavasylyk: I don`t quite understand what the problem is

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,35 @@
...
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final String SPLIT_FOR_DATE = "-";
+    private static final int MIN_VALID_AGE = 35;
+    private static final int MIN_VALID_PERIOD = 10;
+    private static final int DATE_SIZE = 2;
+    private static final int YEAR_FROM;
+    private static final int YEAR_TO;
+    private static String[] date;
```
hronograf, line 13: it would be better to make `date` a local variable

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,35 @@
...
+
+    static {
+        date = new String[DATE_SIZE];
+        YEAR_FROM = 0;
+        YEAR_TO = 1;
+    }
```
hronograf, line 19: looks like we don't need this static block

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,35 @@
...
+        return candidate.getAge() >= MIN_VALID_AGE
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals("Ukrainian")
+                && isValidPeriods(candidate);
+    }
+
+    private boolean isValidPeriods(Candidate candidate) {
+        date = candidate.getPeriodsInUkr().split(SPLIT_FOR_DATE);
+        int yearFrom = Integer.parseInt(date[0]);
+        int yearTo = Integer.parseInt(date[1]);
```
hronograf, line 32: comment about magic numbers wasn't fixed

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +35,12 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(i -> i % 2 != 0
+                ? numbers.get(i) - 1 : numbers.get(i))
+                .filter(i -> i % 2 != 0)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
hronograf, line 43:
```suggestion
        return IntStream.range(0, numbers.size())
            .map(i -> i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i))
            .filter(i -> i % 2 != 0)
            .average()
            .orElseThrow(NoSuchElementException::new);
```

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +71,11 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        WorkablePeopleValidator workablePeopleValidator
+                = new WorkablePeopleValidator(fromAge, maleToAge, femaleToAge);
```
hronograf, line 75: nice, and also you could create it as anonymous class

hronograf (changes requested):

src/main/java/practice/StreamPractice.java
anastasiavasylyk: write you in my own channel in slack. Check this message

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +35,11 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(i -> i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i))
+                .filter(i -> i % 2 != 0)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
hronograf, line 42:
```suggestion
                .orElseThrow();
```

hronograf (approved):
👍

PR #485, opened 2022-10-27
v1

vyshnevsky:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,25 @@
...
+    private static final String DATA_SEPARATOR = "-";
+    private static final int MIN_AGE_VALUE = 35;
+    private static final int MIM_TERM_LIVING_IN_UKRAINE = 10;
+
+    @Override
+    public boolean test(Candidate candidate) {
+        int beforeDataSplit = Integer.parseInt(candidate.getPeriodsInUkr()
+                .split(DATA_SEPARATOR)[INDEX_DATA_BEFORE_SPLIT]);
+        int afterDataSplit = Integer.parseInt(candidate.getPeriodsInUkr()
+                .split(DATA_SEPARATOR)[INDEX_DATA_AFTER_SPLIT]);
```
shmex-j, line 19: It's more effective to use single split

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,25 @@
...
+        return candidate.isAllowedToVote()
+                && candidate.getNationality().equals(NATIONALITY_UKRAINIAN)
```
shmex-j, line 21: It's optional, but i think it's better to call equals on constant, rather than on variable(which can be null)

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +36,15 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return numbers.stream()
+                 .map(n -> numbers.indexOf(n) % 2 == 1
+                         ? n - 1
+                         : n)
+                 .filter(n -> n % 2 == 1)
+                 .distinct()
```
shmex-j, line 44: You should get all numbers, not only unique so don't use distinct

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +36,15 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return numbers.stream()
+                 .map(n -> numbers.indexOf(n) % 2 == 1
+                         ? n - 1
+                         : n)
```
shmex-j, line 42: I think it's optional but you can use IntStream.range(), then you don't need to call indexOf which is additional loop

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +74,12 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex().equals(Person.Sex.MAN)
+                        && p.getAge() >= fromAge && p.getAge() <= maleToAge
+                        || p.getSex().equals(Person.Sex.WOMAN)
+                        && p.getAge() >= fromAge && p.getAge() <= femaleToAge)
```
shmex-j, line 81: For long conditions it's better to create your own Predicate

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +36,15 @@ public int findMinEvenNumber(List<String> numbers) {
...
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return numbers.stream()
+                 .map(n -> numbers.indexOf(n) % 2 == 1
+                         ? n - 1
+                         : n)
+                 .filter(n -> n % 2 == 1)
+                 .distinct()
+                 .mapToDouble(i -> i)
```
shmex-j, line 45: Not sure that you must use this. I think avreage() method returns OptionalDouble even if it's stream of Integers

src/main/java/practice/CandidateValidator.java
vyshnevsky: Agree with you

src/main/java/practice/CandidateValidator.java
vyshnevsky: I'm calling to constant here, or am I wrong?)

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +16,18 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        if (numbers.size() == 0) {
+            throw new RuntimeException("Can't get min value from list");
+        }
```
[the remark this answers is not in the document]
vyshnevsky: thanks, agree with you

src/main/java/practice/StreamPractice.java
vyshnevsky: idk about IntStream.range(), 
thanks, I'll try to do this

src/main/java/practice/StreamPractice.java
vyshnevsky: ok, thanks)

src/main/java/practice/StreamPractice.java
vyshnevsky: agree with you, its a good practice to create your own implementation of Predicate in new Class, but I think, that in this case its not necessary

src/main/java/practice/StreamPractice.java
vyshnevsky: yes, you're right, its true

src/main/java/practice/CandidateValidator.java
shmex-j: Я ма на увазі що краще робити NATIONALITY_UKRAINIAN.equals(candidate.getNationality()) тому, що якщо candidate.getNationality() поверне null то виникне нульпоінтер а NATIONALITY_UKRAINIAN - константа, тож нулем бути не може, але це не критично в рамках цієї задачі, тож можна залишити так.

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +33,22 @@ public int findMinEvenNumber(List<String> numbers) {
...
+                .orElseThrow(NoSuchElementException::new);
+//        return numbers.stream()
+//                 .map(n -> numbers.indexOf(n) % 2 == 1
+//                         ? n - 1
+//                         : n)
+//                 .filter(n -> n % 2 == 1)
+//                 .distinct()
+//                 .mapToDouble(i -> i)
+//                 .average()
+//                 .getAsDouble();
```
shmex-j, line 51: Закоментований код краще прибирати коли він уже не потрібен

shmex-j (changes requested):
Minor change

src/main/java/practice/CandidateValidator.java
vyshnevsky: туплю) зрозумів дякую!

shmex-j (approved):
Great!

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,24 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int INDEX_DATA_BEFORE_SPLIT = 0;
+    private static final int INDEX_DATA_AFTER_SPLIT = 1;
```
hronograf, line 8: naming should be meaningful, like `YEAR_FROM_INDEX` instead of `INDEX_DATA_BEFORE_SPLIT`

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +33,13 @@ public int findMinEvenNumber(List<String> numbers) {
...
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(n -> n % 2 == 0
+                        ? numbers.get(n)
+                        : numbers.get(n) - 1)
+                .filter(i -> i % 2 == 1)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
hronograf, line 42:
```suggestion
                .orElseThrow();
```

src/main/java/practice/StreamPractice.java
hronograf: let's use Predicate here, just for practice

hronograf (changes requested):
Good!

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +33,22 @@ public int findMinEvenNumber(List<String> numbers) {
...
+                .orElseThrow();
+//        return numbers.stream()
+//                 .map(n -> numbers.indexOf(n) % 2 == 1
+//                         ? n - 1
+//                         : n)
+//                 .filter(n -> n % 2 == 1)
+//                 .distinct()
+//                 .mapToDouble(i -> i)
+//                 .average()
+//                 .getAsDouble();
```
hronograf, line 51: let's remove commented code

hronograf (approved):
Nice

PR #486, opened 2022-10-27
Stream_practice completed

vazonchik:


marchTooMuch (approved):
As for me, all is ok

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,74 @@
...
-     * return the average of all odd numbers from the list or throw NoSuchElementException.
-     * But before that subtract 1 from each element on an odd position (having the odd index).
-     */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return numbers.stream()
+                .map(i -> numbers.indexOf(i) % 2 == 0 ? i : i - 1)
+                .filter(i -> i % 2 != 0)
+                .mapToDouble(i -> (double) i)
+                .distinct()
```
hronograf, line 32:
```suggestion
```
we shouldn't have unique values

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,74 @@
...
 
-    /**
-     * Given a List of Integer numbers,
-     * return the average of all odd numbers from the list or throw NoSuchElementException.
-     * But before that subtract 1 from each element on an odd position (having the odd index).
-     */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return numbers.stream()
+                .map(i -> numbers.indexOf(i) % 2 == 0 ? i : i - 1)
```
hronograf, line 29: let's use `IntStream.range` here, because if we have few equal numbers, our program may work incorrectly

hronograf (changes requested):
Good

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,74 @@
...
-     * Given a List of Integer numbers,
-     * return the average of all odd numbers from the list or throw NoSuchElementException.
-     * But before that subtract 1 from each element on an odd position (having the odd index).
-     */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(i -> i % 2 == 0 ? numbers.get(i) : numbers.get(i) - 1)
+                .filter(i -> i % 2 != 0)
+                .mapToDouble(i -> (double) i)
```
hronograf, line 32:
```suggestion
```

hronograf (approved):
🔥

PR #493, opened 2022-10-29
Implemented practice methods

sequencerr:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,111 @@
...
-        return 0D;
+        Supplier<NoSuchElementException> noElementExceptionSupplier = () ->
+                new NoSuchElementException("Can't get avg value from list: " + numbers);
+        return Math.floor(numbers
+                .stream()
+                .mapToInt(i -> i)
+                .map(INCREMENT_VALUE_ON_ODD_POSITION)
+                .filter(i -> !isEven(i))
+                .average()
+                .orElseThrow(noElementExceptionSupplier));
```
Veronika-Tretiak, line 60: orElseThrow())

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,111 @@
...
 public class StreamPractice {
-    /**
-     * Given list of strings where each element contains 1+ numbers:
-     * input = {"5,30,100", "0,22,7", ...}
-     * return min integer value. One more thing - we're interested in even numbers.
-     * If there is no needed data throw RuntimeException with message
-     * "Can't get min value from list: < Here is our input 'numbers' >"
-     */
+    private static final String COMMA_SPLITTER = ",";
+    private static final IntUnaryOperator INCREMENT_VALUE_ON_ODD_POSITION = new IntUnaryOperator() {
```
Veronika-Tretiak, line 19: I think it's optional but why do you use inner classes to implement your task if it's possible to do it in a short way with stream

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,111 @@
...
-    /**
-     * Given a List of Integer numbers,
-     * return the average of all odd numbers from the list or throw NoSuchElementException.
-     * But before that subtract 1 from each element on an odd position (having the odd index).
-     */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        Supplier<NoSuchElementException> noElementExceptionSupplier = () ->
+                new NoSuchElementException("Can't get avg value from list: " + numbers);
+        return Math.floor(numbers
```
Veronika-Tretiak, line 54: It's optional. It's great if your solution is workable but I think we have easier way to implement this task with IntStream.range

src/main/java/practice/StreamPractice.java
sequencerr: Sorry?

Veronika-Tretiak (approved):

ChumakAn (approved):

PR #495, opened 2022-10-29
New branch

ZoreslavT:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,27 @@
...
+                .split(DELIMITER)[END_PERIOD_IN_UKR])
+                - Integer.parseInt(candidate.getPeriodsInUkr()
+                .split(DELIMITER)[START_PERIOD_IN_UKR]) > MIN_PERIOD_IN_UKR
+                && candidate.getAge() >= MIN_CANDIDATE_AGE
+                && candidate.getNationality().equals(VALID_NATIONALITY)
+                && candidate.isAllowedToVote()) {
+            return true;
+        }
+        return false;
+    }
```
AlinaOryshak, line 26: I think it would be better to call `candidate.getPeriodsInUkr().split(DELIMITER)` method only once before condition and then use it's result by indexes
Don't complicate if construction ! You can write all conditions in return statement

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,72 @@
...
-     */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(i -> (i % 2 != 0)
+                        ? numbers.get(i) - NUMBER_TO_SUBTRACT
+                        : numbers.get(i))
+                .filter(i -> i % 2 != 0)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
AlinaOryshak, line 36: image

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,72 @@
...
-     * Example: select people of working age
-     * (from 18 y.o. and to 60 y.o. for men and to 55 y.o. for women inclusively).
-     */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(i -> (i.getSex() == MAN)
+                        ? i.getAge() >= fromAge && i.getAge() <= maleToAge
+                        : i.getAge() >= fromAge && i.getAge() <= femaleToAge)
```
AlinaOryshak, line 52: image

AlinaOryshak (changes requested):
Good job !! Let's improve your solution

src/main/java/practice/StreamPractice.java
ZoreslavT: Using getAsDouble generate warning - [Optional.get()' without 'isPresent()' check], but whatever....

src/main/java/practice/StreamPractice.java
ZoreslavT: As for me using of two function interfases in method  getOddNumAvrage will complicate the code. 
So I want to leave it as it is.

src/main/java/practice/CandidateValidator.java
```diff
@@ -13,15 +13,19 @@ public class CandidateValidator implements Predicate<Candidate> {
...
-                .split(DELIMITER)[END_PERIOD_IN_UKR])
-                - Integer.parseInt(candidate.getPeriodsInUkr()
-                .split(DELIMITER)[START_PERIOD_IN_UKR]) > MIN_PERIOD_IN_UKR
+        if (getPeriodInUkr(candidate) > MIN_PERIOD_IN_UKR
                 && candidate.getAge() >= MIN_CANDIDATE_AGE
                 && candidate.getNationality().equals(VALID_NATIONALITY)
                 && candidate.isAllowedToVote()) {
             return true;
         }
         return false;
```
AlinaOryshak, line 22: What do you think about it ?
```
return getPeriodInUkr(candidate) > MIN_PERIOD_IN_UKR
                && candidate.getAge() >= MIN_CANDIDATE_AGE
                && candidate.getNationality().equals(VALID_NATIONALITY)
                && candidate.isAllowedToVote();
```

AlinaOryshak (changes requested):

AlinaOryshak:
Nice

AlinaOryshak (approved):

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,74 @@
...
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(e -> Arrays.stream(e.split(COMA)))
+                .map(Integer::parseInt)
+                .mapToInt(e -> e)
+                .filter(e -> e % 2 == 0)
+                .min()
+                .orElseThrow(() ->
+                        new RuntimeException("Can't get min value from list: " + numbers));
```
ChumakAn, line 26: it's better to throw NoSuchElementException

ChumakAn (changes requested):
Good! Only minor comment

ChumakAn (approved):

PR #496, opened 2022-10-29
Stream practice

AlinaOryshak:


ZoreslavT (approved):
LGFM!)

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,82 @@
...
-     * "Can't get min value from list: < Here is our input 'numbers' >"
-     */
+    private static final String DELIMITER = ",";
+    private static final int SUBTRAHEND = 1;
+
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(s -> s.split(DELIMITER))
+                .flatMap(Arrays::stream)
```
ChumakAn, line 20: we can replace it with one operation
```suggestion
                .flatMap(s -> Arrays.stream(s.split(DELIMETER)))
```

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,82 @@
...
-     * and live in Ukraine for 10 years. For the last requirement use field periodsInUkr,
-     * which has following view: "2002-2015"
-     * We want to reuse our validation in future, so let's write our own impl of Predicate
-     * parametrized with Candidate in CandidateValidator.
-     */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        CandidateValidator validator = new CandidateValidator();
+        return candidates.stream()
+                .filter(validator)
```
ChumakAn, line 77:
```suggestion
        return candidates.stream()
                .filter(new CandidateValidator())
```

ChumakAn (changes requested):
Good! Only minor comments

ChumakAn (approved):

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,80 @@
...
+    private static final int SUBTRAHEND = 1;
+
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(s -> Arrays.stream(s.split(DELIMITER)))
+                .mapToInt(Integer::parseInt)
+                .filter(n -> n % 2 == 0)
+                .min()
+                .orElseThrow(() -> new RuntimeException("Can't get min value from list: "
```
ChumakAn, line 23: Also, it would be a good practice to use NoSuchElementException

PR #498, opened 2022-10-29
hw-stream-practice

Iryna-a:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,79 @@
...
+
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(n -> n.split(REGEX))
+                .flatMap(Arrays::stream)
+                .map(Integer::parseInt)
+                .filter(i -> i % 2 == 0)
+                .min(Integer::compareTo)
+                .orElseThrow(() -> new RuntimeException("Can't get min value from list: "
```
OleksiyMasaltsev, line 20: If you use constant for regex, I guess it is a good idea to use it with the message as well.

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,79 @@
...
-     * If there is no needed data throw RuntimeException with message
-     * "Can't get min value from list: < Here is our input 'numbers' >"
-     */
+    private static final String REGEX = ",";
+
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(n -> n.split(REGEX))
+                .flatMap(Arrays::stream)
```
OleksiyMasaltsev, line 16: You can skip this map method call using flatMap immidiately:
`.flatMap(s -> Arrays.stream(s.split(REGEX)))`

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,79 @@
...
-     */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return numbers.stream()
+                .map(n -> numbers.indexOf(n) % 2 != 0 ? n - 1 : n)
+                .filter(n -> n % 2 != 0)
+                .mapToInt(n -> n)
+                .distinct()
+                .average()
+                .getAsDouble();
```
OleksiyMasaltsev, line 31: > API Note:
The preferred alternative to this method is orElseThrow().

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,79 @@
...
-     * or to `femaleToAge` (for women) inclusively.
-     * <p>
-     * Example: select people of working age
-     * (from 18 y.o. and to 60 y.o. for men and to 55 y.o. for women inclusively).
-     */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> getPersonByAgeAndSex(p, fromAge, femaleToAge, maleToAge))
```
OleksiyMasaltsev, line 44: Try to use Predicate interface as a filter parameter.

> Use validator properly
[https://mate-academy.github.io/jv-program-common-mistakes/java-core/java-eight-part-two/stream-api#use-validator-properly](url)

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,79 @@
...
+        if (person.getSex() == Person.Sex.WOMAN && person.getAge() >= fromAge
+                && person.getAge() <= femaleToAge) {
+            return true;
+        }
+        if (person.getSex() == Person.Sex.MAN && person.getAge() >= fromAge
+                && person.getAge() <= maleToAge) {
+            return true;
+        }
+        return false;
     }
```
OleksiyMasaltsev, line 59: It would look better as Predicate with some lambdas.

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,79 @@
...
-     * have nationality - 'Ukrainian'
-     * and live in Ukraine for 10 years. For the last requirement use field periodsInUkr,
-     * which has following view: "2002-2015"
-     * We want to reuse our validation in future, so let's write our own impl of Predicate
-     * parametrized with Candidate in CandidateValidator.
-     */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(c -> new CandidateValidator().test(c))
```
OleksiyMasaltsev, line 74: You can use just `new CandidateValidator()` as parameter here.

OleksiyMasaltsev (changes requested):
Nice work!

src/main/java/practice/StreamPractice.java
Iryna-a: I think getAsDouble() is better here. It also can throw NoSuchElementException

OleksiyMasaltsev (approved):
Super!

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,69 @@
...
+    private static final String MESSAGE = "Can't get min value from list: ";
+
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(n -> Arrays.stream(n.split(REGEX)))
+                .map(Integer::parseInt)
+                .filter(i -> i % 2 == 0)
+                .min(Integer::compareTo)
+                .orElseThrow(() -> new RuntimeException(MESSAGE + numbers));
```
KostyaNikiforov, line 21: It is better to throw NSEE instead of RE

KostyaNikiforov (approved):
🔥

PR #499, opened 2022-10-30
Create test method for candidates and implement other methods

Mar4uk21:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,67 @@
...
-     * select from the List only men whose age is from `fromAge` to `toAge` inclusively.
-     * <p>
-     * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
-     */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(e -> e.getSex() == Person.Sex.MAN)
+                .filter(e -> e.getAge() > fromAge && e.getAge() <= toAge)
+                .collect(Collectors.toList());
```
luka-lozovyk, line 37: It is said in the task that uou should take ages INCUSIVELY. If you have an 18-year-old person who can be recruited to army, such case will not pass your method., because you start with 19 years. Use '>=' instead '>'.
Also, it is a better way to compare enum values by equals(), as ling as if smth is wrong with an enum, you will never get  NullPoiterException from '==' comparison

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,67 @@
...
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(e -> e.getAge() >= fromAge)
+                .filter(e -> e.getAge() <= maleToAge
+                        && e.getSex() == Person.Sex.MAN
+                        || e.getAge() <= femaleToAge
+                        && e.getSex() == Person.Sex.WOMAN)
+                .collect(Collectors.toList());
```
luka-lozovyk, line 48: U can extract predicate conditions into predicate declaratioand and pass it s a parameter to .filter(.....)

luka-lozovyk (changes requested):

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,67 @@
...
-     */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(new CandidateValidator())
+                .map(Candidate::getName)
+                .sorted()
+                .collect(Collectors.toList());
     }
 }
```
luka-lozovyk, line 67: Looks good!

luka-lozovyk (approved):

src/main/java/practice/StreamPractice.java
Mar4uk21: Thx

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,67 @@
...
-     * return the average of all odd numbers from the list or throw NoSuchElementException.
-     * But before that subtract 1 from each element on an odd position (having the odd index).
-     */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(e -> e % 2 == 0 ? numbers.get(e) : numbers.get(e) - 1)
+                .filter(e -> e % 2 == 1)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
KostyaNikiforov, line 30: This exception is default 
```suggestion
                .orElseThrow();
```

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,67 @@
...
-     * new Person(«Helen», 42, Sex.WOMAN))`,
-     * select from the List only men whose age is from `fromAge` to `toAge` inclusively.
-     * <p>
-     * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
-     */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(e -> e.getSex().equals(Person.Sex.MAN))
+                .filter(e -> e.getAge() >= fromAge && e.getAge() <= toAge)
```
KostyaNikiforov, line 37: Let's reduce the number of iterations by combining two filter operations into one. Also, you can use predicate, private methods or boolean var to keep your code readable

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):

PR #500, opened 2022-10-30
task complete, candidateValidator impl

yehbreus:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,67 @@
...
-     * input = {"5,30,100", "0,22,7", ...}
-     * return min integer value. One more thing - we're interested in even numbers.
-     * If there is no needed data throw RuntimeException with message
-     * "Can't get min value from list: < Here is our input 'numbers' >"
-     */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers
+                .stream()
+                .map(s -> List.of(s.split(",")))
```
Veronika-Tretiak, line 16: lets make magic splitter "," constant

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,67 @@
...
+                .flatMap(Collection::stream)
+                .mapToInt(Integer::parseInt)
+                .filter(e -> e % 2 == 0)
+                .min().orElseThrow(() -> new RuntimeException("Can't get min value from list"));
```
Veronika-Tretiak, line 20: .min()
.orElseThrow(() -> new RuntimeException("Can't get min value from list"));

"Can't get min value from list: < Here is our input 'numbers' >"

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,67 @@
...
-     */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(i -> i % 2 == 0
+                        ? numbers.get(i)
+                        : numbers.get(i) - 1)
+                .filter(i -> i % 2 != 0)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
Veronika-Tretiak, line 30: .orElseThrow();
by default you will get NoSuchElementException

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,67 @@
...
-     * (from 18 y.o. and to 60 y.o. for men and to 55 y.o. for women inclusively).
-     */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(h -> h.getAge() >= fromAge
+                        && ((h.getSex() == Person.Sex.MAN && h.getAge() <= maleToAge)
+                        || (h.getSex() == Person.Sex.WOMAN && h.getAge() <= femaleToAge)))
+                .collect(Collectors.toList());
```
Veronika-Tretiak, line 47: for such long condition its better to make Predicate with this condition

Veronika-Tretiak (changes requested):
Great job!
Let's fix a few mistakes

src/main/java/practice/StreamPractice.java
yehbreus: I don't really see the need for it. Let's leave this question to the mentor

Veronika-Tretiak (approved):

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,69 @@
...
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers
+                .stream()
+                .map(s -> List.of(s.split(SPLITTER)))
+                .flatMap(Collection::stream)
+                .mapToInt(Integer::parseInt)
+                .filter(e -> e % 2 == 0)
+                .min()
+                .orElseThrow(() -> new RuntimeException("Can't get min value from list" + numbers));
```
KostyaNikiforov, line 22: It is better to throw NSEE in such cases

src/main/java/practice/StreamPractice.java
KostyaNikiforov: For example, you can use Predicate of boolean local vars to make your code more readable. Like, you separate describing of condition with a stream filtering. I found it as a kind of SRP inside a method

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):

PR #503, opened 2022-10-30
released streamPractice and CandidateValidator

apc10bbe:


Iryna-a (approved):
LGTM!

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +50,11 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getAge() >= fromAge
```
ChumakAn, line 54: what about using a Predicate?

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +69,11 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getAge() >= fromAge
+                        && ((p.getSex() == Person.Sex.MAN && p.getAge() <= maleToAge)
+                        || (p.getSex() == Person.Sex.WOMAN && p.getAge() <= femaleToAge)))
```
ChumakAn, line 75: same here

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +103,11 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        CandidateValidator candidateValidator = new CandidateValidator();
+        return candidates.stream()
+                .filter(candidateValidator)
```
ChumakAn, line 108:
```suggestion
        return candidates.stream()
                .filter(new CandidateValidator())
```

ChumakAn (changes requested):
Good! Minor comments

src/main/java/practice/StreamPractice.java
ChumakAn: *Optional

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(n -> Arrays.stream(n.split(SPLITTER)))
+                .mapToInt(Integer::parseInt)
+                .filter(i -> i % 2 == 0)
+                .min()
+                .orElseThrow(() -> new RuntimeException("Can't get min value from list: "
```
ChumakAn, line 27: it's better no throw NoSUchElementException

apc10bbe: year, but this condition at task, that we should throw RE with message:
"1. Given list of strings where each element contains 1+ numbers: input = {"5,30,100", "0, 22, 7", ...} return min integer value. One more thing - we're interested in even numbers. If there is no needed data throw RuntimeException with message "Can't get min value from list: method_input_list"."

ChumakAn: it's a child one of RuntimeException 
image

ChumakAn (changes requested):

ChumakAn (approved):

PR #507, opened 2022-10-31
Released Stream API

OleksandrKorniychuk:


Veronika-Tretiak (approved):

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,28 @@
...
+        return candidate.getAge() >= ADMISSIBLE_AGE
+                && candidate.isAllowedToVote()
+                && getValue(candidate.getPeriodsInUkr())
+                && candidate.getNationality().equals(NATIONALITY);
+    }
+
+    private boolean getValue(String period) {
+        final String [] years = period.split(HYPHEN);
+        final int givenPeriod;
+        givenPeriod = Integer.parseInt(years[INDEX_AFTER]) - Integer.parseInt(years[INDEX_BEFORE]);
```
KostyaNikiforov, line 25:
```suggestion
        final int givenPeriod = Integer.parseInt(years[INDEX_AFTER]) - Integer.parseInt(years[INDEX_BEFORE]);
```

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,13 @@ public class StreamPractice {
...
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(n -> n.split(COMMA))
+                .flatMap(Arrays::stream)
+                .mapToInt(Integer::parseInt)
+                .filter(n -> n % 2 == 0)
+                .min()
+                .orElseThrow(() -> new RuntimeException(RUNTIME_MESSAGE_OUT_EXCEPTION));
```
KostyaNikiforov, line 29: It is better to throw NSEE in such cases instead of RE

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):

PR #509, opened 2022-10-31
Hw jv stream GitHub practice solution

romangolovatyi:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,13 @@ public class StreamPractice {
...
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(n -> n.split(","))
+                .flatMap(Arrays::stream)
+                .mapToInt(Integer::parseInt)
+                .filter(n -> n % 2 == 0)
+                .min()
+                .orElseThrow(() -> new NoSuchElementException("Can't get min value from list"));
```
Veronika-Tretiak, line 27: .orElseThrow(() -> new NoSuchElementException("Can't get min value from list" + numbers));

 "Can't get min value from list: < Here is our input 'numbers' >"

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +33,13 @@ public int findMinEvenNumber(List<String> numbers) {
...
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(n -> n % 2 == 0
+                        ? numbers.get(n)
+                        : numbers.get(n) - 1)
+                .filter(i -> i % 2 == 1)
+                .average()
+                .orElseThrow(() -> new NoSuchElementException("Can't get average value from list"));
```
Veronika-Tretiak, line 42: orElseThrow();
you will get NoSuchElementException by default

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +69,12 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex().equals(Person.Sex.MAN)
+                        && p.getAge() >= fromAge && p.getAge() <= maleToAge
+                        || p.getSex().equals(Person.Sex.WOMAN)
+                        && p.getAge() >= fromAge && p.getAge() <= femaleToAge)
```
Veronika-Tretiak, line 76: for such long condition it is better to make Predicate with this condition

Veronika-Tretiak (changes requested):

src/main/java/practice/StreamPractice.java
romangolovatyi: My carelessness. Thank you!

src/main/java/practice/StreamPractice.java
romangolovatyi: Indeed, it looks better if shortened

src/main/java/practice/StreamPractice.java
romangolovatyi: Thank you for you remarks and spend time

Veronika-Tretiak (approved):

PR #511, opened 2022-11-01
Commit 1

dima-senko:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,22 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MIN_AGE_FOR_CANDIDATES = 35;
```
zachar-chorny, line 7:
```suggestion
    private static final int MIN_AGE = 35;
```
This can be so unspecified, as it is the only constant for minimum age

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,22 @@
...
+    private static final int MIN_YEARS_LIVE_IN_UKRAINE = 10;
+    private static final int INDEX_BEGIN_PERIOD_IN_UKRAINE = 0;
+    private static final int INDEX_LAST_PERIOD_IN_UKRAINE = 1;
```
zachar-chorny, line 10: same

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,22 @@
...
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] split = candidate.getPeriodsInUkr().split("-");
```
zachar-chorny, line 14:
```suggestion
        String[] periods = candidate.getPeriodsInUkr().split("-");
```

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,17 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        OptionalInt min = numbers.stream()
+                .map(x -> Arrays.stream(x.split(","))
+                        .collect(Collectors.toList()))
+                .flatMap(Collection::stream)
```
zachar-chorny, line 26:
```suggestion
                .map(s -> s.split(","))
                .flatMap(Arrays::stream)
```
It can be simplified like this)

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,17 @@ public class StreamPractice {
...
+                .mapToInt(Integer::parseInt)
+                .filter(x -> x % 2 == 0)
```
zachar-chorny, line 28: x is not informative name in this case

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,17 @@ public class StreamPractice {
...
+                .min();
+        if (min == OptionalInt.empty()) {
+            throw new RuntimeException("Can't get min value from list: " + numbers);
+        }
```
zachar-chorny, line 32: Lets add .orElseThrow after min instead of this if

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,17 @@ public class StreamPractice {
...
+        return min.getAsInt();
```
zachar-chorny, line 33: use return instead of OptionalInt

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +113,10 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(x -> new CandidateValidator().test(x))
+                .map(x -> x.getName())
```
zachar-chorny, line 118: Replace x everywhere, with more informative letters. Usually use the first letter of the objects in the stream

zachar-chorny (changes requested):

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +83,12 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex() == Person.Sex.WOMAN && p.getAge() >= femaleAge)
+                .map(Person::getCats)
+                .flatMap(Collection::stream)
```
aaananas, line 89:
```suggestion
                .flatMap(person -> person.getCats().stream())
```

aaananas (changes requested):

Rommelua (approved):

PR #512, opened 2022-11-03
Version 1.Done

SerhiiRovenskyi:


src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +107,10 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(candidate -> new CandidateValidator().test(candidate))
```
aaananas, line 111:
```suggestion
                .filter(new CandidateValidator())
```

aaananas (changes requested):

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +17,13 @@ public class StreamPractice {
...
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(s -> Arrays.stream(s.split(",")))
+                .map(Integer::parseInt)
+                .filter(integer -> integer % 2 == 0)
+                .reduce(Integer::min)
+                .orElseThrow(() -> new RuntimeException("Can't get min value from list: "
+                        + "< Here is our input 'numbers' >"));
```
zachar-chorny, line 26:
```suggestion
        return numbers.stream()
                .flatMap(s -> Arrays.stream(s.split(",")))
                .mapToInt(Integer::parseInt)
                .filter(integer -> integer % 2 == 0)
                .min()
                .orElseThrow(() -> new RuntimeException("Can't get min value from list: "
                        + "< Here is our input 'numbers' >"));
```

zachar-chorny, line 26: "< Here is our input 'numbers' >"? where?)

zachar-chorny (changes requested):

aaananas (approved):

PR #515, opened 2022-11-13
solution was added

kateryna-mykh:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,21 @@
...
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MIN_AGE = 35;
+    private static final String NATIONALITY = "Ukrainian";
+    private static final int LIVING_PERIOD = 10;
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] periodInUk = candidate.getPeriodsInUkr().split("-");
+        int currentLivingPeriod = Integer.parseInt(periodInUk[1])
+                - Integer.parseInt(periodInUk[0]);
```
kozhukhovsky, line 15: you can create constants for 1 and 0

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,107 @@
...
-import java.util.Collections;
+import java.util.Arrays;
 import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import model.Candidate;
 import model.Person;
 
 public class StreamPractice {
+
```
kozhukhovsky, line 11: remove empty line

kozhukhovsky (approved):

src/main/java/practice/CandidateValidator.java
kateryna-mykh: fixes were made

src/main/java/practice/StreamPractice.java
kateryna-mykh: fixes were made

PR #518, opened 2022-11-18
Hm second branch

SofiiaKazanivska:
All necessary methods were implemented using Stream.  + empty line, renamed variable, changed method getCatsNames.( created new branch)


aleksandr-bevza (approved):
Looks good. But variable names like "periodsInUk" are confusing. Because UK - is known as United Kingdom, but Ukraine is UKR

PR #519, opened 2022-11-18
1. added implementation to methods: findMinEvenNumber(), getOddNumsAv…

MTSKL:
…erage(), selectMenByAge();

2. created method checkedPerson(), which checks the input parameters and returns a boolean value and added implementation getWorkablePeople()
3. added implementation to method getCatsNames();
4. checked the candidates by parameters in the CandidateValidator class, where the test method of the Predicate interface was overridden and added implementation to method validateCandidates();

### DIFF
--- a/src/main/java/practice/CandidateValidator.java
+++ b/src/main/java/practice/CandidateValidator.java
@@ -1,5 +1,22 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MIN_AGE = 35;
+    private static final String NATIONALITY = "Ukrainian";
+    private static final int MIN_YEAR_IN_UA = 10;
+
+    @Override
+    public boolean test(Candidate candidate) {
+        if (candidate.getAge() < MIN_AGE
+                || !candidate.isAllowedToVote()
+                || !candidate.getNationality().equals(NATIONALITY)) {
+            return false;
+        }
+        String[] candidatePeriodsOfUA = candidate.getPeriodsInUkr().split("-");
+        return (Integer.parseInt(candidatePeriodsOfUA[1])
+                - Integer.parseInt(candidatePeriodsOfUA[0])) >= MIN_YEAR_IN_UA;
+    }
 }
--- a/src/main/java/practice/StreamPractice.java
+++ b/src/main/java/practice/StreamPractice.java
@@ -1,11 +1,17 @@
 package practice;
 
-import java.util.Collections;
+import java.util.Collection;
 import java.util.List;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import model.Candidate;
+import model.Cat;
 import model.Person;
 
 public class StreamPractice {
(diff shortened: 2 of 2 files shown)


src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +88,13 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex().equals(Person.Sex.WOMAN)
+                        && p.getAge() >= femaleAge)
+                .map(Person::getCats)
+                .flatMap(Collection::stream)
```
aleksandr-bevza, line 95:
```suggestion
                .flatMap(person -> person.getCats().stream())
```

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,22 @@
...
+    @Override
+    public boolean test(Candidate candidate) {
+        if (candidate.getAge() < MIN_AGE
+                || !candidate.isAllowedToVote()
+                || !candidate.getNationality().equals(NATIONALITY)) {
+            return false;
+        }
+        String[] candidatePeriodsOfUA = candidate.getPeriodsInUkr().split("-");
+        return (Integer.parseInt(candidatePeriodsOfUA[1])
+                - Integer.parseInt(candidatePeriodsOfUA[0])) >= MIN_YEAR_IN_UA;
```
aleksandr-bevza, line 20: Move delimiter "-", index "0" and index "1" to constants

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,22 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MIN_AGE = 35;
+    private static final String NATIONALITY = "Ukrainian";
```
aleksandr-bevza, line 8:
```suggestion
    private static final String REQUIRED_NATIONALITY = "Ukrainian";
```

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +77,9 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> CheckedPerson.check(p, fromAge, femaleToAge, maleToAge))
+                .collect(Collectors.toList());
```
aleksandr-bevza, line 82: It's not a good idea to use inner class in this place. 
"filter()" consumes Predicate. It's better to implement Predicate here.                
```suggestion
       Predicate <Person> personPredicate = ... ;
                return peopleList
                    .stream()
                    .filter(personPredicate)
                    .collect(Collectors.toList());
```

aleksandr-bevza (changes requested):

zachar-chorny (approved):

PR #525, opened 2022-11-25
First commit

Vasyl-Piznak:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,16 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
```
aleksandr-bevza, line 21: This method should be reworked.
1. This task can be done with just a single stream. Start this method with:
return numbers.stream().
2. flatMap() would be very helpful in this stream
3. throw exception at the end of stream if there is no even numbers in it

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +37,15 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        if (numbers.isEmpty()) {
+            throw new NoSuchElementException();
+        }
```
aleksandr-bevza, line 42: This can be deleted. NoSuchElementException will be thrown automatically at the end of stream if there is nothing to return
```suggestion     
```

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,22 @@
...
+        int[] calculate = new int[2];
+        String [] years = candidate.getPeriodsInUkr().split("-");
+        for (int i = 0; i < years.length; i++) {
+            calculate[i] = Integer.parseInt(years[i]);
+        }
+        int peroid = calculate[1] - calculate[0];
+        return candidate.getAge() >= 35
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals("Ukrainian")
+                && peroid >= 10;
```
aleksandr-bevza, line 19: Please move 35, Ukrainian, 10 to constants

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,22 @@
...
+public class CandidateValidator implements Predicate<Candidate> {
+
+    @Override
+    public boolean test(Candidate candidate) {
+        int[] calculate = new int[2];
+        String [] years = candidate.getPeriodsInUkr().split("-");
+        for (int i = 0; i < years.length; i++) {
+            calculate[i] = Integer.parseInt(years[i]);
+        }
+        int peroid = calculate[1] - calculate[0];
```
aleksandr-bevza, line 15: Here are just 2 indexes. No need to use loop here. It makes code difficult to read
```suggestion
        String [] years = candidate.getPeriodsInUkr().split("-");
        int firstYear = Integer.parseInt(years[0]);
        int lastYear = Integer.parseInt(years[1]);
        int peroid = lastYear - firstYear ;
```

aleksandr-bevza (changes requested):

aleksandr-bevza (approved):

PR #529, opened 2022-12-08
practice is done

OlexiiMelnik:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,23 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int START_YEAR = 2002;
+    private static final int FINISH_YEAR = 2015;
```
zachar-chorny, line 8: Why are you using your data to check? That's not the way it's supposed to be! Everything you need to check the candidate is in their fields

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,23 @@
...
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] split = candidate.getPeriodsInUkr().split("-");
+        int first = Integer.valueOf(split[0]);
+        int second = Integer.valueOf(split[1]);
+        if (candidate.getAge() >= 35 && candidate.isAllowedToVote()
+                && candidate.getNationality().equals("Ukrainian")
+                && second - first >= 10
+                && (first >= START_YEAR || second <= FINISH_YEAR)) {
+            return true;
```
zachar-chorny, line 19: Create constants for magic numbers and "Ukrainian" string

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(s -> s.split(","))
+                .flatMap(s -> Arrays.stream(s))
```
zachar-chorny, line 24: Combine map and flatMap

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,13 @@ public class StreamPractice {
...
+                .mapToInt(s -> Integer.valueOf(s))
+                .filter(i -> i % 2 == 0)
+                .min().orElseThrow(() ->
+                        new RuntimeException("Can't get min value from list: method_input_list"));
```
zachar-chorny, line 28:
```suggestion
                        new RuntimeException("Can't get min value from list:" + numbers));
```

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +34,11 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(index -> index % 2 != 0 ? numbers.get(index) - 1 : numbers.get(index))
+                .filter(value -> value % 2 != 0)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
zachar-chorny, line 41: Add an informative message for exclusion

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +83,12 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex().equals(Person.Sex.WOMAN) && p.getAge() >= femaleAge)
+                .map(p -> p.getCats())
+                .flatMap(Collection::stream)
```
zachar-chorny, line 89: Combine map and flatMap

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +104,11 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        CandidateValidator candidateValidator = new CandidateValidator();
+        return candidates.stream()
+                .filter(c -> candidateValidator.test(c))
+                 .map(c -> c.getName())
+                 .sorted()
+                 .collect(Collectors.toList());
```
zachar-chorny, line 112:
```suggestion
        return candidates.stream()
                .filter(new CandidateValidator())
                 .map(c -> c.getName())
                 .sorted()
                 .collect(Collectors.toList());
```

zachar-chorny (changes requested):

src/main/java/practice/CandidateValidator.java
OlexiiMelnik: I compare it with the condition.

kozhukhovsky: It is better not to compare, but to calculate the period by subtracting two dates and then comparing it

src/main/java/practice/CandidateValidator.java
kozhukhovsky: not all fixed

kozhukhovsky (changes requested):
Not all fixed

aleksandr-bevza (approved):

PR #542, opened 2022-12-25
stream solution added

Uka0001:


src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +71,13 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getAge() >= 18)
```
Dmytro-Mk, line 75: you should use fromAge past in method and hardcode in '18'

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +71,13 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
...
+                .filter(p -> p.getSex()
+                        .equals(Person.Sex.MAN)
+                        ? p.getAge() <= maleToAge
+                        : p.getAge() <= femaleToAge)
```
Dmytro-Mk, line 79: we don't need second .filter(), you can put all the logic in one

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,26 @@
...
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+
+    public static final String VALID_NATIONALITY = "Ukrainian";
+    public static final int PERIOD_TO = 1;
+    public static final int PERIOD_FROM = 0;
+    public static final int VALID_PERIOD = 10;
+    public static final int VALID_AGE = 35;
```
Dmytro-Mk, line 12:
```suggestion
    public static final String REQUIRED_NATIONALITY = "Ukrainian";
    public static final int YEAR_TO_INDEX = 1;
    public static final int YEAR_FROM_INDEX = 0;
    public static final int REQUIRED_MIN_PERIOD = 10;
    public static final int REQUIRED_MIN_AGE = 35;
```

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,26 @@
...
+    public boolean test(Candidate candidate) {
+        return candidate.isAllowedToVote()
+                && candidate.getNationality().equals(VALID_NATIONALITY)
+                && candidate.getAge() >= VALID_AGE
+                && Integer.parseInt(candidate
+                .getPeriodsInUkr()
+                .split("-")[PERIOD_TO])
+                - Integer.parseInt(candidate
+                .getPeriodsInUkr()
+                .split("-")[PERIOD_FROM]) >= VALID_PERIOD;
```
Dmytro-Mk, line 24: to increase readability let's move this logic to a new local variable or separate method

Dmytro-Mk (changes requested):

src/main/java/practice/StreamPractice.java
Uka0001: fixed

src/main/java/practice/StreamPractice.java
Uka0001: fixed

src/main/java/practice/CandidateValidator.java
Uka0001: fixed

src/main/java/practice/CandidateValidator.java
Uka0001: fixed

Dmytro-Mk (approved):

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,29 @@
...
+                && countPeriodInUkraine(candidate) >= REQUIRED_MIN_PERIOD;
+    }
+
+    private static int countPeriodInUkraine(Candidate candidate) {
+        return Integer.parseInt(candidate
+                .getPeriodsInUkr()
+                .split("-")[YEAR_TO_INDEX])
+                - Integer.parseInt(candidate
+                .getPeriodsInUkr()
+                .split("-")[YEAR_FROM_INDEX]);
```
KostyaNikiforov, line 27: Let's call method split only once in this method

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +34,12 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream
+                .range(0, numbers.size())
+                .map(i -> i % 2 == 1 ? numbers.get(i) - 1 : numbers.get(i))
+                .filter(i -> i % 2 == 1)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
KostyaNikiforov, line 42: This exception is default 
```suggestion
                .orElseThrow();
```

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +107,11 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        CandidateValidator candidateValidator = new CandidateValidator();
+        return candidates.stream()
+                .filter(candidateValidator)
+                .map(Candidate::getName)
```
KostyaNikiforov, line 113:
```suggestion
        return candidates.stream()
                .filter(new CandidateValidator())
                .map(Candidate::getName)
```

KostyaNikiforov (changes requested):

src/main/java/practice/CandidateValidator.java
Uka0001: fixed

src/main/java/practice/StreamPractice.java
Uka0001: fixed

src/main/java/practice/StreamPractice.java
Uka0001: fixed

KostyaNikiforov (approved):

PR #549, opened 2023-01-01
StreamPractice was implemented.

katrevichdv:
StreamPractice, CandidateValidator were implemented. IsEven, PersonSexAndAgePredicate were added.


src/main/java/practice/IsEven.java
```diff
@@ -0,0 +1,11 @@
+package practice;
+
+import java.util.function.IntPredicate;
+
+public class IsEven implements IntPredicate {
+
+    @Override
+    public boolean test(int value) {
+        return (value & 1) == 0;
+    }
```
karma-o, line 10: we don't need a whole class just to test if the number is even

src/main/java/practice/PersonSexAndAgePredicate.java
```diff
@@ -0,0 +1,23 @@
+package practice;
+
+import java.util.function.Predicate;
+import model.Person;
+
+public class PersonSexAndAgePredicate implements Predicate<Person> {
```
karma-o, line 6: this is also redundant, if you don't want to write big or complex operations inside streams, you can move it into a local variable, except the candidate validator, since it is part of the task.
This is not really a mistake, so you can leave it as it is, because you are actually using it multiple times with different parameters, but I would say that in most cases we have no need for a separate class for a stream operation

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +22,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(s -> s.split(SEPARATOR))
+                .flatMap(Arrays::stream)
```
karma-o, line 27: image
```suggestion
                .flatMap(s -> Arrays.stream(s.split(SEPARATOR)))
```
check similar cases

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +87,14 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        Predicate<Person> predicate =
```
karma-o, line 90: can we come up with better naming?

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +87,14 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
...
+                new PersonSexAndAgePredicate(Person.Sex.WOMAN, femaleAge, Integer.MAX_VALUE);
+        return peopleList.stream()
+                .filter(predicate)
+                .map(Person::getCats)
+                .flatMap(List::stream)
```
karma-o, line 95: image

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +110,10 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(new CandidateValidator())
```
karma-o, line 114: move candidate validator to a local variable

karma-o (changes requested):

src/main/java/practice/IsEven.java
katrevichdv: I really don't see the difference between IsEven and CandidateValidator, they are both stateless. So, cauz one of predicates is separete class I moved 2 others to separete classes.

src/main/java/practice/StreamPractice.java
katrevichdv: It's redundant. It's stateless, used only once and this solution is allowed.
[https://mate-academy.github.io/jv-program-common-mistakes/java-core/java-eight-part-two/stream-api#use-validator-properly](url)

aaananas (approved):

PR #551, opened 2023-01-02
all works

diana-lomei:


src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +84,12 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex().equals(Person.Sex.WOMAN) && p.getAge() >= femaleAge)
+                .map(Person::getCats)
+                .flatMap(Collection::stream)
```
aleksandr-bevza, line 90: You can do it in one line
```suggestion
                .flatMap(person -> person.getCats().stream())
```

aleksandr-bevza (changes requested):

src/main/java/practice/CandidateValidator.java
```diff
@@ -4,6 +4,8 @@
 import model.Candidate;
 
 public class CandidateValidator implements Predicate<Candidate> {
+    private static int ZERO_INDEX = 0;
+    private static int FIRST_INDEX = 1;
```
aleksandr-bevza, line 8: These index names must be informative. Not just zero and first. For example:
LIVES_IN_UKRAINE_TO_INDEX,  LIVES_IN_UKRAINE_FROM_INDEX

aleksandr-bevza (changes requested):

aleksandr-bevza (approved):

PR #554, opened 2023-01-05
Solved stream

o-shyshkan:


src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +91,12 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(s -> s.getAge() >= femaleAge && s.getSex() == Person.Sex.WOMAN)
+                .map(s -> s.getCats())
+                .flatMap(s -> s.stream())
```
aleksandr-bevza, line 97: You can do it in one line
```suggestion
                .flatMap(person -> person.getCats().stream())
```

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +112,11 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        Predicate<Candidate> predicate = new CandidateValidator();
+        return candidates.stream()
+                .filter(predicate::test)
```
aleksandr-bevza, line 117: No need to call method "test" here. It will be called automatically because this is Predicate
```suggestion
                .filter(new CandidateValidator())
```

aleksandr-bevza (changes requested):

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,23 @@
...
-public class CandidateValidator {
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
     //write your code here
+    private static final int MIN_TERM_IN_UKR = 35;
+    private static final int START_YEAR_INDEX = 0;
+    private static final int END_YEAR_INDEX = 1;
+    private static final int TEN_YEARS = 10;
```
okuzan, line 11: use constants to give the description for the values, it's clear that value is 10. I think you meant actually MIN_TERM_IN_UKR

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,23 @@
 package practice;
 
-public class CandidateValidator {
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
     //write your code here
+    private static final int MIN_TERM_IN_UKR = 35;
```
okuzan, line 8:
```suggestion
    private static final int MIN_AGE = 35;
```

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +33,11 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream
+               .range(0, numbers.size())
```
okuzan, line 37:
```suggestion
        return IntStream.range(0, numbers.size())
```

okuzan (changes requested):
naming issues

okuzan (approved):
good job!

PR #561, opened 2023-01-08
implemented task:

yurii-dikalo:
all tests have passed;
mvn clean package - build success.

### DIFF
--- a/src/main/java/practice/CandidateValidator.java
+++ b/src/main/java/practice/CandidateValidator.java
@@ -1,5 +1,27 @@
 package practice;
 
-public class CandidateValidator {
+import java.util.Arrays;
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MIN_AGE = 35;
+    private static final int FROM_YEAR_INDEX = 0;
+    private static final int TO_YEAR_INDEX = 1;
+    private static final int MIN_YEARS_IN_UKRAINE = 10;
+    private static final String DELIMITER = "-";
+    private static final String NATIONALITY = "Ukrainian";
+
+    @Override
+    public boolean test(Candidate candidate) {
+        int [] periods = Arrays.stream(candidate.getPeriodsInUkr().split(DELIMITER))
+                .mapToInt(Integer::parseInt)
+                .toArray();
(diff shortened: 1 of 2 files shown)


Avassor3000 (approved):
pr is closely near to ideal

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,27 @@
...
+    private static final int TO_YEAR_INDEX = 1;
+    private static final int MIN_YEARS_IN_UKRAINE = 10;
+    private static final String DELIMITER = "-";
+    private static final String NATIONALITY = "Ukrainian";
+
+    @Override
+    public boolean test(Candidate candidate) {
+        int [] periods = Arrays.stream(candidate.getPeriodsInUkr().split(DELIMITER))
+                .mapToInt(Integer::parseInt)
+                .toArray();
```
KostyaNikiforov, line 19:
```suggestion
        int[] periods = candidate.getPeriodsInUkr().split(DELIMITER);
```
)

KostyaNikiforov (changes requested):

src/main/java/practice/CandidateValidator.java
yurii-dikalo: this returns String[] which causes compile error... also it was the exact reason I had to use this workaround. though I do understand that there were ways to do this without using Stream API I didn't want to make any of listed common mistakes

I could do this
String[] periodStrings = candidate.getPeriodsInUkr().split(DELIMITER);
int yearsInUkr = Integer.parseInt(periodStrings [TO_YEAR_INDEX]) - Integer.parseInt(periodStrings [FROM_YEAR_INDEX]);

do you want me to do it this way?

yurii-dikalo: ok, whatever, I redid it anyway

KostyaNikiforov (approved):

src/main/java/practice/CandidateValidator.java
KostyaNikiforov: You are right about compile error. You need to transform data anyway. But for me, this option looks better because you don't create new redundant objects, like stream and array.

yurii-dikalo: oh, I see it now, thank you for your comment!

PR #573, opened 2023-01-10
Hw

lysytskapolina:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,66 @@
...
-     * "Can't get min value from list: < Here is our input 'numbers' >"
-     */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(s -> Arrays.asList(s.split(",")))
+                .flatMap(Collection::stream)
+                .map(Integer::parseInt)
+                .filter(i -> i % 2 == 0)
+                .min(Comparator.naturalOrder())
```
OleksiiZaloznyi, line 20:
```suggestion
                .mapToInt(Integer::parseInt)
                .filter(i -> i % 2 == 0)
                .min()
```
We can select MapToInt here and use the .min() method without a comparator

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,26 @@
...
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MIN_VALID_AGE = 35;
+    private static final String VALID_NATIONALITY = "Ukrainian";
+    private static final int MIN_VALID_PERIOD = 10;
+    private static final int FROM_YEAR_INDEX = 0;
+    private static final int TO_YEAR_INDEX = 1;
```
OleksiiZaloznyi, line 11:
```suggestion
    private static final int MIN_VALID_AGE = 35;
    private static final int MIN_VALID_PERIOD = 10;
    private static final int FROM_YEAR_INDEX = 0;
    private static final int TO_YEAR_INDEX = 1;
    private static final String VALID_NATIONALITY = "Ukrainian";
```

I think you can order constants by type. I'm not sure about this exactly, but it looks more readable

OleksiiZaloznyi (changes requested):

OleksiiZaloznyi (approved):

OleksiiZaloznyi (approved):

KostyaNikiforov (approved):

KostyaNikiforov (approved):

PR #578, opened 2023-01-10
validator was created

leonid-topal:


src/main/java/practice/StreamPractice.java
```diff
@@ -42,7 +41,7 @@ public Double getOddNumsAverage(List<Integer> numbers) {
                 .mapToDouble(n -> n)
                 .distinct()
                 .average()
-                .getAsDouble();
+                .orElseThrow(NoSuchElementException::new);
```
DimaSmyk, line 44: You can use .orElseThrow() without creating NoSuchException. Because this method generates NoSuchException by default

DimaSmyk (approved):

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +103,10 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(n -> new CandidateValidator().test(n))
```
KostyaNikiforov, line 107:
```suggestion
                .filter(new CandidateValidator())
```

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):

PR #582, opened 2023-01-13
1. All tasks have been done.

Taurusden:
2. In last task method CandidateValidator.test has been created, which determine candidate accordingly to conditions for president post.

### DIFF
--- a/src/main/java/practice/CandidateValidator.java
+++ b/src/main/java/practice/CandidateValidator.java
@@ -1,5 +1,22 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int PERIOD_InUKR = 10;
+    private static final int CANDIDATE_ValidAGE = 35;
+    private static final String VALID_NATIONALITY = "Ukrainian";
+
+    @Override
+    public boolean test(Candidate candidate) {
+        return getPeriodsInUkr(candidate.getPeriodsInUkr()) > PERIOD_InUKR
+                && candidate.isAllowedToVote() && candidate.getAge() >= CANDIDATE_ValidAGE
+                && candidate.getNationality().equals(VALID_NATIONALITY);
+    }
+
+    private int getPeriodsInUkr(String getPeriodsInUkr) {
+        String[] period = getPeriodsInUkr.split("-");
+        return Integer.parseInt(period[1]) - Integer.parseInt(period[0]);
+    }
 }
--- a/src/main/java/practice/StreamPractice.java
+++ b/src/main/java/practice/StreamPractice.java
@@ -1,29 +1,39 @@
 package practice;
 
-import java.util.Collections;
+import java.util.Arrays;
+import java.util.Collection;
 import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import model.Candidate;
+import model.Cat;
 import model.Person;
 
-public class StreamPractice {
-    /**
-     * Given list of strings where each element contains 1+ numbers:
-     * input = {"5,30,100", "0,22,7", ...}
-     * return min integer value. One more thing - we're interested in even numbers.
-     * If there is no needed data throw RuntimeException with message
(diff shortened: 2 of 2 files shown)


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,22 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int PERIOD_InUKR = 10;
+    private static final int CANDIDATE_ValidAGE = 35;
```
okuzan, line 8: you just invented some new case? 😅 
Constants should be written with capital letters in Java, where the words are separated by an underscore ('_') character.

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,22 @@
...
+    @Override
+    public boolean test(Candidate candidate) {
+        return getPeriodsInUkr(candidate.getPeriodsInUkr()) > PERIOD_InUKR
+                && candidate.isAllowedToVote() && candidate.getAge() >= CANDIDATE_ValidAGE
+                && candidate.getNationality().equals(VALID_NATIONALITY);
+    }
+
+    private int getPeriodsInUkr(String getPeriodsInUkr) {
+        String[] period = getPeriodsInUkr.split("-");
+        return Integer.parseInt(period[1]) - Integer.parseInt(period[0]);
```
okuzan, line 20: array indexes should also be constants, with descriptive names

src/main/java/practice/CandidateValidator.java
okuzan, line 8:
```suggestion
    private static final int MIN_AGE = 35;
```

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,22 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int PERIOD_InUKR = 10;
```
```suggestion
    private static final int MIN_YEARS_IN_UKR= 10;
```

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,22 @@
...
+
+    @Override
+    public boolean test(Candidate candidate) {
+        return getPeriodsInUkr(candidate.getPeriodsInUkr()) > PERIOD_InUKR
+                && candidate.isAllowedToVote() && candidate.getAge() >= CANDIDATE_ValidAGE
+                && candidate.getNationality().equals(VALID_NATIONALITY);
+    }
+
+    private int getPeriodsInUkr(String getPeriodsInUkr) {
+        String[] period = getPeriodsInUkr.split("-");
```
okuzan, line 19: string should be constant

okuzan (changes requested):

src/main/java/practice/CandidateValidator.java
Taurusden: CheckStile was sleepnig).

src/main/java/practice/CandidateValidator.java
Taurusden: Don't understand, that getPeriodsInUkr is different each time, we get it from each candidate.

Taurusden: Есть, друг подсказал). Не понял что сепаратор тоже надо).

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,25 @@
...
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MIN_YEARS_IN_UKR = 10;
+    private static final int MIN_AGE = 35;
+    private static final String VALID_NATIONALITY = "Ukrainian";
+    private static final int START_LIVING = 0;
```
okuzan, line 10:
```suggestion
    private static final int STARTED_LIVING_INDEX = 0;
```

okuzan (approved):
Good job

Taurusden:
Thanks.

PR #586, opened 2023-01-17
Added solutions for tasks from "StreamPractice"

vgladkyi:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,15 @@ public class StreamPractice {
...
-        return 0;
+        OptionalInt min = numbers
+                .stream()
+                .map(el -> el.split(","))
+                .flatMap(Arrays::stream)
+                .mapToInt(Integer::valueOf)
+                .filter(el -> el % 2 == 0)
+                .min();
+        return min.orElseThrow(() -> new RuntimeException("Can't get min value from list: "
+                + numbers));
```
nacenik, line 30:
```suggestion
        return numbers
                .stream()
                .map(el -> el.split(","))
                .flatMap(Arrays::stream)
                .mapToInt(Integer::parseInt)
                .filter(el -> el % 2 == 0)
                .min()
                .orElseThrow(() -> new RuntimeException("Can't get min value from list: "
                + numbers));
```

nacenik: parseInt is better here, cause it returns int, in that time when valueOf returns wrapper Integer

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +52,11 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList
+                .stream()
+                .filter(person -> person.getSex().equals(Person.Sex.MAN))
+                .filter(person -> person.getAge() >= fromAge && person.getAge() <= toAge)
```
nacenik, line 58:
```suggestion
                .filter(person -> person.getSex().equals(Person.Sex.MAN)
                && person.getAge() >= fromAge && person.getAge() <= toAge)
```

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +87,12 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList
+                .stream()
+                .filter(person -> person.getSex().equals(Person.Sex.WOMAN))
+                .filter(person -> person.getAge() > femaleAge)
```
nacenik, line 93:
```suggestion
                .filter(person -> person.getSex().equals(Person.Sex.WOMAN)
                && person -> person.getAge() > femaleAge))
```

nacenik (changes requested):

src/main/java/practice/StreamPractice.java
vgladkyi: Thank you for feedback!

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +70,14 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
...
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getSex().equals(Person.Sex.WOMAN)
+                        && person.getAge() >= fromAge
+                        && person.getAge() <= femaleToAge
+                        || person.getSex().equals(Person.Sex.MAN)
+                        && person.getAge() >= fromAge
+                        && person.getAge() <= maleToAge)
```
Ivan95kos, line 79: If you have a long boolean expression you may create your own implementation of Predicate and use it in your filter.

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,15 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers
+                .stream()
```
Ivan95kos, line 22: It is better to leave the stream() at the same level as the variable for which we are calling
```suggestion
        return numbers.stream()
```

Ivan95kos (approved):
Great job

PR #587, opened 2023-01-20
STREAM PRACTICE COMPLETED

viktoriia-maherko:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,20 +1,35 @@
...
      * return min integer value. One more thing - we're interested in even numbers.
      * If there is no needed data throw RuntimeException with message
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
+
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(s -> s.split(","))
+                .flatMap(Arrays::stream)
```
```suggestion
                .flatMap(s -> s.split(",").stream())
```
use just 1 map, instead of multiple if possible

aaananas (changes requested):

aaananas (approved):

PR #588, opened 2023-01-22
Solved given tasks using stream API, wrote logic for CandidateValidator

nkharyshyn:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,23 @@
...
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final String UA_NATIONALITY = "Ukrainian";
+    private static final int INDEX_YEAR_LIVE_FROM = 0;
+    private static final int INDEX_YEAR_LIVE_TO = 1;
+    private static final int NECESSARY_PERIOD_IN_UA = 10;
```
nacenik, line 10:
```suggestion
    private static final int MIN_PERIOD_IN_UA = 10;
```

nacenik (changes requested):

okuzan (approved):
Good job!

PR #591, opened 2023-01-26
feat: jv-stream-practice #comment stream practice

sania517:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,20 @@
...
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MIN_AGE = 35;
+    private static final int MIN_PERIOD_IN_UKRAINE = 10;
+    private static final String MANDATORY_NATIONALITY = "Ukrainian";
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] years = candidate.getPeriodsInUkr().split("-");
```
okuzan, line 13: "-" should be constant

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,20 @@
...
+        int period = Integer.parseInt(years[1]) - Integer.parseInt(years[0]);
```
okuzan, line 14: indexes should be constants

okuzan (changes requested):

okuzan (approved):
Good job!

PR #600, opened 2023-02-05
Hw jv stream GitHub practice

yaaariiik:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,29 @@
...
+    private static final Integer REQUIRED_PERIOD = 10;
+    private static final String SEPARATOR_SYMBOL = "-";
+    private static final int FROM_DATE_INDEX = 0;
+    private static final int TO_DATE_INDEX = 1;
+
+    private final Function<String, Integer> getPeriod = s -> {
+        String[] separatedDates = s.split(SEPARATOR_SYMBOL);
+        return Integer.parseInt(separatedDates[TO_DATE_INDEX])
+                - Integer.parseInt(separatedDates[FROM_DATE_INDEX]);
+    };
```
andriy-shulepa, line 19: I don't see a reason to make it lambda. refactor to plain private method, please

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,74 @@
...
-    /**
-     * Given list of strings where each element contains 1+ numbers:
-     * input = {"5,30,100", "0,22,7", ...}
-     * return min integer value. One more thing - we're interested in even numbers.
-     * If there is no needed data throw RuntimeException with message
-     * "Can't get min value from list: < Here is our input 'numbers' >"
-     */
+    private static final String SPLITER = ",";
+    private static final int CHECK_EVEN_NUMBER = 2;
+    private static final int SUBTRACT_NUMBER = 1;
```
andriy-shulepa, line 17: In general, it's a good practice to extract numbers to constant but in this case, it makes code harder to read.
x % 2 == 0 - it's a common expression to check if a number is even, so it can be left as is
(i - 1) - is also common expression to get previous index and can be left as is

andriy-shulepa (changes requested):

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,71 @@
...
-     * "Can't get min value from list: < Here is our input 'numbers' >"
-     */
+    private static final String SPLITER = ",";
+
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(s -> Arrays.asList(s.split(SPLITER)))
+                .flatMap(Collection::stream)
+                .mapToInt(Integer::valueOf)
```
nacenik, line 21: valueOf - return wrapper
parseInt - return primitive
```suggestion
                .mapToInt(Integer::parseInt)
```

nacenik (approved):
Good job!

PR #601, opened 2023-02-05
Solution of task Stream practice. All tests are done

stepanpeknyy:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,16 @@ public class StreamPractice {
...
+        try {
+            return numbers.stream()
+                    .flatMap(e -> Arrays.stream(e.split(",")))
+                    .map(e -> Integer.valueOf(e))
+                    .filter(i -> i % 2 == 0)
+                    .min(Integer:: compare)
+                    .get();
+        } catch (RuntimeException e) {
+            throw new RuntimeException("Can't get min value from list");
+        }
```
nacenik, line 31:
```suggestion
        return numbers.stream()
                .flatMap(e -> Arrays.stream(e.split(",")))
                .map(e -> Integer.valueOf(e))
                .filter(i -> i % 2 == 0)
                .min(Integer:: compare)
                .orElseThrow(() -> new RuntimeException("Can't get min value from list: " + numbers));
```

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +37,25 @@ public int findMinEvenNumber(List<String> numbers) {
...
+                .filter(e -> e % 2 != 0)
+                .boxed()
+                .collect(Collectors.toList());
+        List<Integer> listOddAfterSubtraction = IntStream.range(0, numbers.size())
+                .filter(i -> i % 2 != 0)
+                .mapToObj(i -> numbers.get(i) - 1)
+                .mapToInt(Integer::intValue)
+                .filter(e -> e % 2 != 0)
+                .boxed()
+                .collect(Collectors.toList());
```
nacenik, line 53: you don't need 2 this streams. Use only one stream and use ternary operation in map `.map(i -> i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i))`

src/main/java/practice/WorkablePeopleValidator.java
```diff
@@ -0,0 +1,16 @@
+package practice;
+
+import java.util.function.Predicate;
+import model.Person;
+
+public class WorkablePeopleValidator implements Predicate<Person> {
```
nacenik, line 6: you don't need this class

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +85,10 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        WorkablePeopleValidator workablePeopleValidator = new WorkablePeopleValidator();
```
nacenik, line 88: you can create an anonymous Predicate class instead of a real class

nacenik (changes requested):

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,21 @@
...
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MIN_AGE = 35;
+    private static final int MIN_TIME_IN_UKR = 10;
+
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= MIN_AGE && candidate.isAllowedToVote()
+                && candidate.getNationality().equals("Ukrainian")
+                && Arrays.stream(candidate.getPeriodsInUkr().split("-"))
+                .map(Integer::valueOf)
```
okuzan, line 16:
```suggestion
```

okuzan (changes requested):

PR #602, opened 2023-02-05
Hw jv stream practice

Iskamele:
The task for getOddNumsAverage is miswritten since my, and not only, Predicate isOdd allows us to pass tests. However, it works with odd numbers, and not with odd indices of these numbers in the array, as stated in the conditions of the task.

Or maybe I'm just confused by now.

### DIFF
--- a/src/main/java/practice/StreamPractice.java
+++ b/src/main/java/practice/StreamPractice.java
@@ -1,16 +1,14 @@
 package practice;
 
-import model.Candidate;
-import model.Cat;
-import model.Person;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.List;
-import java.util.NoSuchElementException;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
+import model.Candidate;
+import model.Cat;
+import model.Person;
 
 public class StreamPractice {
     private final Predicate<Integer> isOdd = num -> num % 2 != 0;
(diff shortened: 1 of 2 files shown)


src/main/java/practice/StreamPractice.java
```diff
@@ -1,11 +1,18 @@
...
 import java.util.List;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import model.Candidate;
+import model.Cat;
 import model.Person;
 
 public class StreamPractice {
+    private final Predicate<Integer> isOdd = num -> num % 2 != 0;
```
Rommelua, line 14: You use this only in one method. Make it local or simply use lambda in the stream.

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +85,17 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
...
-        return Collections.emptyList();
+        Predicate<Person> isCatOwnerFemale = person ->
+                person.getSex() == Person.Sex.WOMAN
+                        && person.getAge() >= femaleAge
+                        && person.getCats() != null;
+
+        return peopleList.stream()
+                .filter(isCatOwnerFemale)
+                .map(Person::getCats)
+                .flatMap(Collection::stream)
```
Rommelua, line 96:
```suggestion
                .flatMap(p -> p.getCatList().stream())
```

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +111,25 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
...
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(new CandidateValidator())
+                .map(Candidate::getName)
+                .sorted()
+                .collect(Collectors.toList());
+    }
+
+    private boolean isWorkablePerson(int fromAge, int femaleToAge,
```
Rommelua, line 121: Better create local predicat in getWorkablePeople method

Rommelua (changes requested):

Rommelua (approved):

PR #603, opened 2023-02-06
implement solution

dreeeamcatcher:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,24 @@
...
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] periodInUkraine = candidate.getPeriodsInUkr().split("-");
+        boolean inUkraineForTenYears =
+                Integer.parseInt(periodInUkraine[INDEX_PERIOD_IN_UKRAINE_END])
+                        - Integer.parseInt(periodInUkraine[INDEX_PERIOD_IN_UKRAINE_START])
+                        > 10;
+
+        return candidate.getAge() >= 35
```
aaananas, line 19: lets add constants for required values

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,14 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        String exceptionMessage = "Can't get min value from list: ";
+        return numbers.stream()
+                .map(s -> s.split(","))
+                .flatMap(Arrays::stream)
```
aaananas, line 24:
```suggestion
                .flatMap(s -> Arrays.stream(s.split(",")))
```

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +34,23 @@ public int findMinEvenNumber(List<String> numbers) {
...
+                    if (oddPositionElements.contains(x)) {
+                        oddPositionElements.remove(x);
+                        return x - 1;
+                    }
+                    return x;
+                })
+                .filter(x -> x % 2 != 0)
+                .mapToDouble(x -> x)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
aaananas, line 53:
```
IntStream.range(0, numbers.size())
                .map(i -> i % 2 != 0 ? numbers.get(i) : numbers.get(i) - 1)
                ...```
 lets combine 2 streams
add message to the exception

**Review by aaananas — CHANGES_REQUESTED** [review]

File: src/main/java/practice/StreamPractice.java [hunk 4]
```diff
@@ -14,7 +18,13 @@ public class StreamPractice {
...
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        String exceptionMessage = "Can't get min value from list: ";
+        return numbers.stream()
+                .flatMap(s -> Arrays.stream(s.split(",")))
+                .map(Integer::parseInt)
+                .filter(x -> x % 2 == 0)
+                .min(Integer::compareTo)
+                .orElseThrow(() -> new RuntimeException(exceptionMessage + numbers));
```
```suggestion
        return numbers.stream()
                .flatMap(s -> Arrays.stream(s.split(",")))
                .map(Integer::parseInt)
                .filter(x -> x % 2 == 0)
                .min(Integer::compareTo)
                .orElseThrow(() -> new RuntimeException("Can't get min value from list: " + numbers));
```
storing messages in variables is not a good idea - it is hard to read, and not informative

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +33,13 @@ public int findMinEvenNumber(List<String> numbers) {
...
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        String exceptionMessage = "Can't get an average value with given conditions";
+        return IntStream.range(0, numbers.size())
+                .map(i -> i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i))
+                .filter(x -> x % 2 != 0)
+                .mapToDouble(x -> x)
+                .average()
+                .orElseThrow(() -> new NoSuchElementException(exceptionMessage));
```
aaananas, line 42:
```suggestion
        String exceptionMessage = "Can't get an average value with given conditions";
        return IntStream.range(0, numbers.size())
                .map(i -> i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i))
                .filter(x -> x % 2 != 0)
                .mapToDouble(x -> x)
                .average()
                .orElseThrow(() -> new NoSuchElementException("Can't get an average value with given conditions for list: " + numbers));
```

aaananas (approved):

src/main/java/practice/StreamPractice.java
dreeeamcatcher: agree. but it fails checkstyle compliance because the "line is longer than 100 characters" :man_shrugging:

PR #604, opened 2023-02-08
jv-stream-github-practice solution was created.

mrGvvD:


src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +33,12 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0,numbers.size())
```
kozhukhovsky, line 36:
```suggestion
        return IntStream.range(0, numbers.size())
```

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +33,12 @@ public int findMinEvenNumber(List<String> numbers) {
...
+                .mapToObj(n -> (n % 2 == 0) ? numbers.get(n) : numbers.get(n) - 1)
+                .filter(n -> n % 2 == 1)
+                . mapToDouble(n -> (double) n)
+                .average()
+                .getAsDouble();
```
kozhukhovsky, line 41:
```suggestion
                .map(n -> (n % 2 == 0) ? numbers.get(n) : numbers.get(n) - 1)
                .filter(n -> n % 2 == 1)
                .average()
                .getAsDouble();
```

kozhukhovsky (approved):

PR #622, opened 2023-02-23
StreamPractice method was implemented and CandidateValidator was impl…

GalubaGSV:
…emented

### DIFF
--- a/src/main/java/practice/StreamPractice.java
+++ b/src/main/java/practice/StreamPractice.java
@@ -1,7 +1,13 @@
 package practice;
 
-import java.util.Collections;
+import java.util.Arrays;
 import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.Optional;
+import java.util.OptionalDouble;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import model.Candidate;
 import model.Person;
 
@@ -14,7 +20,16 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        Optional<Integer> min = numbers.stream()
+                .map(n -> n.split(","))
+                .flatMap(n -> Arrays.stream(n))
+                .map(n -> Integer.parseInt(n))
+                .filter(n -> n % 2 == 0)
+                .min((x, y) -> x - y);
+        if (min.isEmpty()) {
+            throw new RuntimeException("Can't get min value from list: " + numbers);
+        }
+        return min.get();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +112,11 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        Predicate<Candidate> predicate = new CandidateValidator();
+        return candidates.stream()
+                .sorted((x, y) -> x.getName().compareTo(y.getName()))
```
mishok321, line 117:
```suggestion
                .sorted(Comparator.comparing(Candidate::getName)))
```
same with method reference

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +112,11 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
...
+                .filter(predicate)
+                .map(c -> c.getName())
```
mishok321, line 119:
```suggestion
                .map(Candidate::getName)
```

mishok321 (changes requested):

mishok321 (approved):

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +22,14 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(n -> n.split(NUMBERS_DELIMITER))
+                .flatMap(Arrays::stream)
```
petro-ushchuk, line 27: You can inline it

```suggestion
                .flatMap(line -> Stream.of(line.split(NUMBERS_DELIMITER)))
```

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +22,14 @@ public class StreamPractice {
...
+                .map(Integer::parseInt)
```
petro-ushchuk, line 28: Call `mapToInt` instead

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +55,12 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getAge()
+                        >= fromAge && p.getAge()
+                        <= toAge && p.getSex()
+                        == Person.Sex.MAN)
```
petro-ushchuk, line 62: Better to extract this condition in separate private method and call method reference.

https://www.baeldung.com/java-streams-multiple-filters-vs-condition

petro-ushchuk (approved):
Good work! There are some possible improvments in comments, but anyway approved. 🔥

PR #633, opened 2023-03-06
Successful stream practice

IlkoGood:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,21 @@
...
+    private static final int MINIMUM_AGE = 35;
+    private static final String REQUIRED_NATIONALITY = "Ukrainian";
+    private static final String DATA_SEPARATOR = "-";
+    private static final int MINIMUM_PERIOD = 10;
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] citizenYears = candidate.getPeriodsInUkr().split(DATA_SEPARATOR);
+        return candidate.getAge() >= MINIMUM_AGE
+                && candidate.getNationality().equals(REQUIRED_NATIONALITY)
```
MaksymHorielikov, line 16:
```suggestion
                && REQUIRED_NATIONALITY.equals(candidate.getNationality())
```

MaksymHorielikov: this option is better, REQUIRED_NATIONALITY is always != null

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,68 @@
...
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(str -> List.of(str.split(DATA_SEPARATOR_TASK_1)))
+                .flatMap(List::stream)
+                .map(Integer::valueOf)
+                .filter(i -> i % 2 == 0)
+                .min(Integer::compare)
+                .orElseThrow(() -> new RuntimeException(
+                 "Can't get min value from list" + numbers));
```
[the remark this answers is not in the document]
MaksymHorielikov: you can reduce intermediate operations, and work with streams of primitives, which will be faster

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,68 @@
...
-     * return the average of all odd numbers from the list or throw NoSuchElementException.
-     * But before that subtract 1 from each element on an odd position (having the odd index).
-     */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .boxed()
+                .map(i -> i % 2 == 0 ? numbers.get(i) : numbers.get(i) - 1)
+                .filter(i -> i % 2 != 0)
+                .mapToInt(Integer::intValue).average().orElseThrow();
```
MaksymHorielikov, line 31:
```suggestion
                .mapToInt(Integer::intValue).average().orElseThrow(NoSuchElementException::new);
```

MaksymHorielikov: according to the conditions of the task, you need to throw NoSuchElementException

IlkoGood: NoSuchElementException is default for orElseThrow.
You can call it directly, but that is not something you really need to.

src/main/java/practice/StreamPractice.java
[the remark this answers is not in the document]
IlkoGood: Well, you are calling stream inside stream, which is kind of "intermediate operation".
So it's not such a diferrence, if we count operations.
But yes, your solutions may be faster.

src/main/java/practice/StreamPractice.java
MaksymHorielikov: thanks, now I will know what exception is thrown in the orElseThrow() method without parameters

src/main/java/practice/CandidateValidator.java
MaksymHorielikov: My counterarguments:
1. My version is safer and doesn't require additional checking of what getPriodsInUkr() returns.
2. Correct processing of incorrect data that may fall into the method parameters is our responsibility.
3. I find it inappropriate to throw exceptions in this class. The validator must check the correctness of the data, if the data is incorrect, return false.
4. All of the above is only a recommendation.

IlkoGood: Yes, throwing an exception was definitely strange move, I agree.

MaksymHorielikov (approved):

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,68 @@
...
-     * If there is no needed data throw RuntimeException with message
-     * "Can't get min value from list: < Here is our input 'numbers' >"
-     */
+    private static final String DATA_SEPARATOR_TASK_1 = ",";
+
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(str -> List.of(str.split(DATA_SEPARATOR_TASK_1)))
+                .flatMap(List::stream)
```
KseniiaMakarova, line 18: 1. let's call only flatMap method
2. why do we need to make a list from array here? it is redundant

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,68 @@
...
+                .mapToInt(Integer::parseInt)
+                .filter(i -> i % 2 == 0)
+                .min()
+                .orElseThrow(() -> new RuntimeException(
+                 "Can't get min value from list" + numbers));
```
KseniiaMakarova, line 23:
```suggestion
                 "Can't get min value from list " + numbers));
```

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,68 @@
...
 
-    /**
-     * Given a List of Integer numbers,
-     * return the average of all odd numbers from the list or throw NoSuchElementException.
-     * But before that subtract 1 from each element on an odd position (having the odd index).
-     */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .boxed()
```
KseniiaMakarova, line 28: why do we need boxed()?

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,68 @@
...
+                .map(i -> i % 2 == 0 ? numbers.get(i) : numbers.get(i) - 1)
+                .filter(i -> i % 2 != 0)
+                .mapToInt(Integer::intValue).average().orElseThrow();
```
KseniiaMakarova, line 31:
```suggestion
                .average()
                .orElseThrow();
```

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,68 @@
...
-     * Example: select people of working age
-     * (from 18 y.o. and to 60 y.o. for men and to 55 y.o. for women inclusively).
-     */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getAge() >= fromAge
+                && ((p.getSex() == Person.Sex.MAN && p.getAge() <= maleToAge)
+                || (p.getSex() == Person.Sex.WOMAN && p.getAge() <= femaleToAge)))
```
KseniiaMakarova, line 47:
```suggestion
                && (p.getSex() == Person.Sex.MAN && p.getAge() <= maleToAge
                || p.getSex() == Person.Sex.WOMAN && p.getAge() <= femaleToAge))
```
&& is evaluated before ||

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,68 @@
...
-     * Given a List of `Person` instances (having `name`, `age`, `sex` and `cats` fields,
-     * and each `Cat` having a `name` and `age`),
-     * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
-     */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex() == Person.Sex.WOMAN && p.getAge() >= femaleAge)
+                .map(Person::getCats)
+                .flatMap(Collection::stream)
```
KseniiaMakarova, line 55:
```suggestion
                .flatMap(...)
```

KseniiaMakarova (changes requested):

src/main/java/practice/StreamPractice.java
IlkoGood: I'm not a magician, only learning stuff, thanks)

src/main/java/practice/StreamPractice.java
IlkoGood: hmm, double convertation int->Integer->int...
I missed something in algorithm, I'll check

src/main/java/practice/StreamPractice.java
IlkoGood: I really don't get for now, as in first flatMap...
I take stream of Lists from map and unpacking it with flatMap.
Not sure how to do it with one step, but ok.

src/main/java/practice/StreamPractice.java
IlkoGood: As I see now, you are talking about same thing as Maksim, 
putting .stream() call inside flatMap...
I really can't catch the necessity of this action.

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,66 @@
...
 
 public class StreamPractice {
-    /**
-     * Given list of strings where each element contains 1+ numbers:
-     * input = {"5,30,100", "0,22,7", ...}
-     * return min integer value. One more thing - we're interested in even numbers.
-     * If there is no needed data throw RuntimeException with message
-     * "Can't get min value from list: < Here is our input 'numbers' >"
-     */
+    private static final String DATA_SEPARATOR_TASK_1 = ",";
```
KseniiaMakarova, line 13:
```suggestion
    private static final String DATA_SEPARATOR = ",";
```

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,66 @@
...
-    /**
-     * Given a List of `Person` instances (having `name`, `age`, `sex` and `cats` fields,
-     * and each `Cat` having a `name` and `age`),
-     * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
-     */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex() == Person.Sex.WOMAN && p.getAge() >= femaleAge)
+                .flatMap(person -> person.getCats().stream())
```
KseniiaMakarova, line 53:
```suggestion
                .flatMap(p -> p.getCats().stream())
```

KseniiaMakarova (approved):

src/main/java/practice/StreamPractice.java
IlkoGood: ok

PR #636, opened 2023-03-07
implementation of all methods in the StreamPractice class, implementa…

MaksymHorielikov:
…tion of the predicate in the CandidateValidator class

### DIFF
--- a/src/main/java/practice/CandidateValidator.java
+++ b/src/main/java/practice/CandidateValidator.java
@@ -1,5 +1,39 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MIN_AGE = 35;
+    private static final int MIN_PERIODS_IN_UKR = 10;
+    private static final String NATIONALITY = "Ukrainian";
+    private static final String SEPARATOR_IN_PERIOD = "-";
+
+    /**
+     * Evaluates this predicate on the given argument.
+     *
+     * @param candidate the input argument
+     * @return {@code true} if the input argument matches the predicate,
+     * otherwise {@code false}
+     */
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= MIN_AGE
+                && candidate.isAllowedToVote()
+                && NATIONALITY.equals(candidate.getNationality())
+                && isValidPeriodsInUkr(candidate.getPeriodsInUkr());
+    }
+
+    private boolean isValidPeriodsInUkr(String periodInUkr) {
+        if (periodInUkr == null) {
+            return false;
+        }
+        String[] range = periodInUkr.split(SEPARATOR_IN_PERIOD);
+        try {
+            return Integer.parseInt(range[1])
+                    - Integer.parseInt(range[0]) + 1 >= MIN_PERIODS_IN_UKR;
+        } catch (NumberFormatException e) {
+            return false;
+        }
+    }
 }
--- a/src/main/java/practice/StreamPractice.java
+++ b/src/main/java/practice/StreamPractice.java
@@ -1,8 +1,11 @@
 package practice;
 
-import java.util.Collections;
+import java.util.Arrays;
 import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.stream.Collectors;
 import model.Candidate;
+import model.Cat;
 import model.Person;
 
 public class StreamPractice {
@@ -14,7 +17,12 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(s -> Arrays.stream(s.split(",")))
+                .mapToInt(Integer::parseInt)
+                .filter(n -> n % 2 == 0)
+                .min()
+                .orElseThrow(() -> new RuntimeException("Can't get min value from list"));
     }
 
     /**
@@ -23,7 +31,12 @@ public class StreamPractice {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        final int[] index = {0};
+        return numbers.stream()
+                .mapToInt(n -> index[0]++ % 2 != 0 ? n - 1 : n)
+                .filter(n -> n % 2 != 0)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
     }
 
     /**
@@ -35,7 +48,11 @@ public class StreamPractice {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex() == Person.Sex.MAN
+                        && p.getAge() >= fromAge
+                        && p.getAge() <= toAge)
+                .collect(Collectors.toList());
     }
 
     /**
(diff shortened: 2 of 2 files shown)


IlkoGood (approved):
Good job!

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,40 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MIN_AGE = 35;
+    private static final int MIN_PERIODS_IN_UKR = 10;
```
petro-ushchuk, line 8: Better keep full names in variables, please rename to `MIN_PERIODS_IN_UKRAINE`

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,40 @@
...
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate != null
+                && candidate.getAge() >= MIN_AGE
+                && candidate.isAllowedToVote()
+                && NATIONALITY.equals(candidate.getNationality())
+                && isValidPeriodsInUkr(candidate.getPeriodsInUkr());
+    }
+
+    private boolean isValidPeriodsInUkr(String periodInUkr) {
```
petro-ushchuk, line 28: And methods names as well

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +17,12 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(s -> Arrays.stream(s.split(",")))
```
petro-ushchuk, line 21: `","` should be constant

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +31,12 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        final int[] index = {0};
+        return numbers.stream()
+                .mapToInt(n -> index[0]++ % 2 != 0 ? n - 1 : n)
```
petro-ushchuk, line 36: What a sense to keep index as arrays zero element? Use a single variable, or inline it with `.range(0, names.length)`

petro-ushchuk (changes requested):

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,40 @@
...
+        if (periodInUkr == null) {
+            return false;
+        }
+        String[] range = periodInUkr.split(SEPARATOR_IN_PERIOD);
+        try {
+            return Integer.parseInt(range[1])
+                    - Integer.parseInt(range[0]) + 1 >= MIN_PERIODS_IN_UKRAINE;
+        } catch (NumberFormatException e) {
+            return false;
+        }
```
KseniiaMakarova, line 38: catching RuntimeExceptions is not a good practice
can we throw a custom exception here? or just not catch it?

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +34,12 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        AtomicInteger index = new AtomicInteger();
+        return numbers.stream()
+                .mapToInt(n -> index.getAndIncrement() % 2 != 0 ? n - 1 : n)
+                .filter(n -> n % 2 != 0)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
KseniiaMakarova, line 42:
```suggestion
                .orElseThrow();
```
will this work? NoSuchElementException is the default exception

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +34,12 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        AtomicInteger index = new AtomicInteger();
+        return numbers.stream()
+                .mapToInt(n -> index.getAndIncrement() % 2 != 0 ? n - 1 : n)
```
KseniiaMakarova, line 39: it is an elegant workaround, but let's please use `IntStream.range(0, numbers.size())`

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +70,11 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getAge() >= fromAge
+                        && ((p.getSex() == Person.Sex.MAN && p.getAge() <= maleToAge)
+                        || (p.getSex() == Person.Sex.WOMAN && p.getAge() <= femaleToAge)))
```
KseniiaMakarova, line 76:
```suggestion
                        && (p.getSex() == Person.Sex.MAN && p.getAge() <= maleToAge
                        || p.getSex() == Person.Sex.WOMAN && p.getAge() <= femaleToAge))
```
&& is evaluated before || so let's remove redundant brackets

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +83,12 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex() == Person.Sex.WOMAN && p.getAge() >= femaleAge)
+                .map(Person::getCats)
+                .flatMap(lc -> lc.stream().map(Cat::getName))
```
KseniiaMakarova, line 89:
```suggestion
                .flatMap(p -> p.getCats().stream())
                .map(Cat::getName)
```

KseniiaMakarova (changes requested):
good job!

src/main/java/practice/CandidateValidator.java
MaksymHorielikov: in this particular case we want the validator to do its job for any incoming data, if the period data cannot be recognized as numbers we simply return false. Throwing an exception doesn't make sense here.

src/main/java/practice/StreamPractice.java
MaksymHorielikov: I can do it...
However, I did a performance test and my version using AtomicInteger will always be several times faster than the version using IntStream.range(0, numbers.size()).

src/main/java/practice/CandidateValidator.java
MaksymHorielikov: Changed this method, simplified. If any exception occurs (when the input string does not match the format described in the task), false is returned.

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,37 @@
...
+                && NATIONALITY.equals(candidate.getNationality())
+                && isValidPeriodsInUkraine(candidate.getPeriodsInUkr());
+    }
+
+    private static boolean isValidPeriodsInUkraine(String periodInUkr) {
+        try {
+            String[] range = periodInUkr.split(SEPARATOR_IN_PERIOD);
+            return Integer.parseInt(range[1])
+                    - Integer.parseInt(range[0]) + 1 >= MIN_PERIODS_IN_UKRAINE;
+        } catch (RuntimeException e) {
```
KseniiaMakarova, line 33: it is not correct to use try-catch as if-else

you should 
1. add check for periodInUkr == null
2. add checking parts with regex (to ensure there are only digits there)

please avoid catching RuntimeExceptions

KseniiaMakarova (changes requested):

src/main/java/practice/CandidateValidator.java
MaksymHorielikov: Disagree:
The method will run slower
because checking and applying regular expressions takes extra time.
Using try ... catch avoids unnecessary checks and the method always works correctly.

Performance no longer important?

Andriy-dz (approved):

PR #638, opened 2023-03-07
implemented test method from Predicate, implemented methods using str…

AlexandLes:
…eam API

### DIFF
--- a/src/main/java/practice/CandidateValidator.java
+++ b/src/main/java/practice/CandidateValidator.java
@@ -4,16 +4,21 @@ import java.util.function.Predicate;
 import model.Candidate;
 
 public class CandidateValidator implements Predicate<Candidate> {
+    private static final int SUITABLE_CANDIDATE_AGE = 35;
+    private static final int YEARS_IN_COUNTRY = 10;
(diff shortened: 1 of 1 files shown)


KyslyiViacheslav (approved):
good

src/main/java/practice/StreamPractice.java
```diff
@@ -1,20 +1,32 @@
...
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers
+                .stream()
+                .flatMap(el -> Arrays.stream(el.split(SEPARATOR)))
+                .mapToInt(Integer::valueOf)
+                .filter(number -> number % 2 == 0)
+                .min()
+                .orElseThrow(() -> new RuntimeException("Can't get min value from list: "
+                + numbers));
```
Andriy-dz, line 29: does it make sense to write a string to the array in the message?

Andriy-dz (changes requested):
A very good decision. Well done! Let's fix some problem

petro-ushchuk (approved):
Good work 🔥

PR #642, opened 2023-03-09
solved tasks

KyslyiViacheslav:
implements all method
implements CandidateValidator


Andriy-dz (changes requested):
A very good decision. Well done! Let's fix some problems

Andriy-dz (approved):

PR #648, opened 2023-03-09
Stream Practice home work

V-Vasylchuk:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,24 @@
...
+    private static final int FIRST_ARRAY_ELEMENT = 0;
+    private static final int SECOND_ARRAY_ELEMENT = 1;
+    private static final int MINIMUM_AGE = 35;
+    private static final int MINIMUM_YEAR_IN_UKR = 10;
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] year = candidate.getPeriodsInUkr().split(YEAR_SPLITTER);
+        int yearInUkr = Integer.parseInt(year[SECOND_ARRAY_ELEMENT])
+                                            - Integer.parseInt(year[FIRST_ARRAY_ELEMENT]);
```
IvanZuravlev228, line 18: In my opinion will be better if you split this variable to two, and rename the constants "FIRST_ARRAY_ELEMENT " and "SECOND_ARRAY_ELEMENT " to more understandable name, because your first element is the beginning of the period and your second element is the end

IvanZuravlev228 (changes requested):
Good jod)

IvanZuravlev228 (approved):

karma-o (approved):
Looks good to me 👍

PR #653, opened 2023-03-09
relize of StreamPractice

kromch:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(s -> Arrays.stream(s.split(",")))
```
sanya-1xx8, line 23: "," - should be constant

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +34,13 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        IntStream indexes = IntStream.range(0, numbers.size());
+        return indexes
+                .map(i -> i % 2 == 0 ? numbers.get(i) : numbers.get(i) - 1)
+                .filter(e -> e % 2 == 1)
+                .average()
+                .orElseThrow(() -> new NoSuchElementException(
```
sanya-1xx8, line 42: As for me  (.getAsDouble) is better, let's leave it for the mentor)

sanya-1xx8 (approved):

src/main/java/practice/StreamPractice.java
kromch: i did add this constant

src/main/java/practice/StreamPractice.java
kromch: when i added getAsDouble method
the orElseThrow method stopped working altogether
i guess that this method doing correct work with optional object

sarakhmen: Agree with @sanya-1xx8 , as stated in Common Mistakes, it's better to replace orElseThrow() call with getAsDouble()

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +105,10 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(new CandidateValidator())
+                .sorted(Comparator.comparing(Candidate::getName))
+                .map(Candidate::getName)
```
sarakhmen, line 111:
```suggestion
                .map(Candidate::getName)
                .sorted()
```
I'd recommend something like this

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +72,11 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getAge() >= fromAge && ((p.getAge() <= maleToAge && p.getSex()
+                        == Person.Sex.MAN) || (p.getAge() <= femaleToAge && p.getSex()
+                        == Person.Sex.WOMAN)))
```
sarakhmen, line 78: Think of how you can simplify this filter condition

sarakhmen (changes requested):
Overall good job, but let's improve your solution :)

src/main/java/practice/StreamPractice.java
kromch: amended

Andriy-dz (approved):
Good Job!!

PR #657, opened 2023-03-09
added solution hw-jv-stream-practice

HelenMatus:


src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +35,12 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream
+                .range(0, numbers.size())
+                .map(i -> i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i))
+                .filter(n -> n % 2 != 0)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
ArtemPonomar, line 43: You could add an informative error message

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +85,14 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
...
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex() == Person.Sex.WOMAN
+                        && p.getCats().size() != 0
+                        && p.getAge() >= femaleAge)
+                .map(Person::getCats)
+                .flatMap(Collection::stream)
+                .map(Cat::getName)
```
ArtemPonomar, line 94: You can combine these three lines into one as such:
```suggestion
.flatMap(p -> p.getCats().stream().map(Cat::getName))
```

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +108,10 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .sorted(Comparator.comparing(Candidate::getName))
+                .filter(new CandidateValidator())
```
ArtemPonomar, line 113: It`s not a mistake, but it is better to remove candidates from stream before sorting, so that sort will work faster with smaller amount of candidates.
```suggestion
                .filter(new CandidateValidator())
                .sorted(Comparator.comparing(Candidate::getName))
```

ArtemPonomar (changes requested):
Good job!

ArtemPonomar (approved):

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,24 @@
...
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= MIN_AGE
+                && candidate.isAllowedToVote()
+                && Objects.equals(candidate.getNationality(), NATIONALITY)
+                && countYearsInUkraine(candidate.getPeriodsInUkr()) >= MIN_YEARS_IN_UKRAINE;
+    }
+
+    private int countYearsInUkraine(String years) {
+        String[] splitYears = years.split("-");
```
sarakhmen, line 21: Let's introduce a constant for '-'

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(Pattern.compile(",")::splitAsStream)
```
sarakhmen, line 23: Let's also introduce a constant for ','

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +34,13 @@ public int findMinEvenNumber(List<String> numbers) {
...
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream
+                .range(0, numbers.size())
+                .map(i -> i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i))
+                .filter(n -> n % 2 != 0)
+                .average()
+                .orElseThrow(() -> new NoSuchElementException("No odd numbers in modified list: "
+                        + numbers));
```
sarakhmen, line 43: As stated in common mistakes, let's use getAsDouble() here instead

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +106,10 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(new CandidateValidator())
+                .sorted(Comparator.comparing(Candidate::getName))
+                .map(Candidate::getName)
```
sarakhmen, line 112: I'd recommend simplifying it. If you call map() operation first, then your sorted() method may take no arguments at all

sarakhmen (changes requested):
Good job, but let's improve your solution :)

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,25 @@
...
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= MIN_AGE
+                && candidate.isAllowedToVote()
+                && Objects.equals(candidate.getNationality(), NATIONALITY)
+                && countYearsInUkraine(candidate.getPeriodsInUkr()) >= MIN_YEARS_IN_UKRAINE;
+    }
+
+    private int countYearsInUkraine(String years) {
+        String[] splitYears = years.split(SPLIT_SEPARATOR);
+        return Integer.parseInt(splitYears[1]) - Integer.parseInt(splitYears[0]);
```
sarakhmen, line 23: I forgot to mention, that indexes should also be constants

sarakhmen (approved):
Good job :)

PR #665, opened 2023-03-10
Implement StreamPractice and CandidateValidator which implements func…

AleksandrLts:
…tional interface predicate and override test method

### DIFF
--- a/src/main/java/model/Candidate.java
+++ b/src/main/java/model/Candidate.java
@@ -77,4 +77,15 @@ public class Candidate {
         return Objects.hash(getAge(), getNationality(),
             isAllowedToVote(), getPeriodsInUkr(), getName());
     }
+
+    @Override
+    public String toString() {
+        return "Candidate{"
+                + "age=" + age
+                + ", nationality='" + nationality + '\''
+                + ", allowedToVote=" + allowedToVote
+                + ", periodsInUkr='" + periodsInUkr + '\''
+                + ", name='" + name + '\''
+                + '}';
+    }
 }
--- a/src/main/java/practice/CandidateValidator.java
+++ b/src/main/java/practice/CandidateValidator.java
@@ -1,5 +1,25 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MIN_PERIOD_IN_UKRAINE = 10;
+    private static final int MIN_AGE_ALLOWED = 35;
+    private static final int FROM_YEAR_INDEX = 0;
+    private static final int TO_YEAR_INDEX = 1;
+    private static final String REGEX = "-";
+    private static final String ALLOWED_NATIONALITY = "Ukrainian";
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] periodInUkr = candidate.getPeriodsInUkr().split(REGEX);
+        int fromYearInUkraine = Integer.parseInt(periodInUkr[FROM_YEAR_INDEX]);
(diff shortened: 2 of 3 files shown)


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,25 @@
...
+    private static final int TO_YEAR_INDEX = 1;
+    private static final String REGEX = "-";
+    private static final String ALLOWED_NATIONALITY = "Ukrainian";
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] periodInUkr = candidate.getPeriodsInUkr().split(REGEX);
+        int fromYearInUkraine = Integer.parseInt(periodInUkr[FROM_YEAR_INDEX]);
+        int toYearInUkraine = Integer.parseInt(periodInUkr[TO_YEAR_INDEX]);
+        int candidatePeriodInUkraine = toYearInUkraine - fromYearInUkraine;
```
KosMarch, line 19: in my opinion, there is no need to create so many variables

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,25 @@
...
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MIN_PERIOD_IN_UKRAINE = 10;
+    private static final int MIN_AGE_ALLOWED = 35;
+    private static final int FROM_YEAR_INDEX = 0;
+    private static final int TO_YEAR_INDEX = 1;
```
KosMarch, line 10: do we really need constants for zero and one?

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +103,10 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(new CandidateValidator())
+                .map(Candidate::getName)
```
KosMarch, line 108: you can combine the two

src/main/java/model/Candidate.java
```diff
@@ -77,4 +77,15 @@ public int hashCode() {
...
+    @Override
+    public String toString() {
+        return "Candidate{"
+                + "age=" + age
+                + ", nationality='" + nationality + '\''
+                + ", allowedToVote=" + allowedToVote
+                + ", periodsInUkr='" + periodsInUkr + '\''
+                + ", name='" + name + '\''
+                + '}';
+    }
```
KosMarch, line 90: Why is this necessary in this task?

KosMarch (changes requested):
Great job! let's improve it a bit

src/main/java/model/Candidate.java
AleksandrLts: It isn't necessary, I needed it to check the fault after tests, I can remove this code, if you want

src/main/java/practice/CandidateValidator.java
AleksandrLts: I think it isn't redundant

src/main/java/practice/StreamPractice.java
AleksandrLts: In which way?)

KosMarch: .filter(new CandidateValidator()).map(c -> c.getName())

KosMarch: Oops, my bad.

src/main/java/practice/CandidateValidator.java
AleksandrLts: I left two varibles, because they make code more readable

PR #667, opened 2023-03-10
made implementations for methods by using streams

pechivo:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,20 @@
...
+    private static final int MIN_PERIODS_IN_UKRAINE = 10;
+    private static final int MINIMAL_AGE = 35;
+    private static final String NATIONALITY = "Ukrainian";
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] splitted = candidate.getPeriodsInUkr().split(DELIMITER);
+        return candidate.getAge() >= MINIMAL_AGE
+                && candidate.getNationality().equals(NATIONALITY)
+                && (Integer.parseInt(splitted[1]) - Integer.parseInt(splitted[0]))
```
AleksandrLts, line 17: In my opinion, better to move indexes in informative constants in this case

AleksandrLts (changes requested):
Great work! Let's improve it a bit

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +79,11 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getAge() >= femaleAge && p.getSex().equals(Person.Sex.WOMAN))
```
AleksandrLts, line 83:
```suggestion
                .filter(p -> p.getAge() >= femaleAge && p.getSex() == Person.Sex.WOMAN)
```

AleksandrLts (approved):

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +66,11 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(c -> c.getSex() == Person.Sex.MAN ? c.getAge() >= fromAge
+                        && c.getAge() <= maleToAge : c.getAge() >= fromAge
+                        && c.getAge() <= femaleToAge)
```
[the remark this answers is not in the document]
petro-ushchuk: please, simplify this condition. try to avoid usage of ternary operator. make it in that way `c.getAge() >= fromAge && case for men && case for woman`

src/main/java/practice/StreamPractice.java
petro-ushchuk: no need to use `equals` for enums, just `==`

petro-ushchuk (changes requested):
Good work, but we should improve your solution 😢

petro-ushchuk (approved):
Great work 🔥

PR #675, opened 2023-03-10
Solved

arthursilver111:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +17,14 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(number -> number.split(","))
+                .flatMap(Arrays::stream)
+                .mapToInt(Integer::parseInt)
+                .filter(i -> i % 2 == 0)
```
Ivan95kos, line 24: For the future, do not use abbreviated names
```suggestion
                .filter(number -> number % 2 == 0)
```

Ivan95kos (approved):

PR #676, opened 2023-03-16
Stream practice solution.

MelnychukMyhailo:


okuzan (changes requested):
Good job! Use more meaningful naming for variables used in streams. `i`, `e`, `s` decrease readability

uliana-andrushchuk (approved):

PR #678, opened 2023-03-19
hw-stream-practice-x750l

iamchgi:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,96 @@
...
+            min = numbers.stream()
+                    .flatMapToInt(string -> Arrays.stream(string.split(REGEX_INTEGERS))
+                            .mapToInt(Integer::parseInt))
+                    .filter(i -> i % 2 == 0)
+                    .min()
+                    .getAsInt();
+        } catch (Exception e) {
+            throw new RuntimeException("Can't get min value from list: < Here is our input '"
+                    + numbers + "' >");
+        }
```
olekskov, line 29: Please use .orElseThrow() with required exception insted of try/catch block

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,96 @@
...
+        try {
+            average = numbers.stream()
+                    .map(incOnOddIndex)
+                    .filter(i -> i % 2 == 1)
+                    .mapToInt(Integer::valueOf)
+                    .average()
+                    .getAsDouble();
+        } catch (Exception e) {
+            throw new NoSuchElementException("Error empty List of numbers");
+        }
```
olekskov, line 50: .orElseThrow()

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,96 @@
...
-     */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        Double average;
+        final Integer[] index = {0};
+        Function<Integer, Integer> incOnOddIndex = integer -> {
+            integer = index[0] % 2 == 1 ? integer - 1 : integer;
+            index[0]++;
+            return integer;
+        };
```
olekskov, line 40: Solution will be more easy to read if you use IntStream.range(0, numbers.size()) and then map it to values list contain at position of value in stream (with required decrement of course)

PR #686, opened 2023-03-23
done 1.0

Gagadu:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,75 @@
...
-     * If there is no needed data throw RuntimeException with message
-     * "Can't get min value from list: < Here is our input 'numbers' >"
-     */
+    private static final int CHECK_FOR_EVEN = 2;
+
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers
+                .stream()
+                .map(str -> List.of(str.split(",")))
```
kozhukhovsky, line 18: it’s not necessary to go to a sheet here, streaming is also perfectly done from an array

kozhukhovsky (approved):

PR #697, opened 2023-04-07
StreamAPI_Solution 

SerhiUA:
Realized all tasks. Used Stream API

### DIFF
--- a/src/main/java/practice/CandidateValidator.java
+++ b/src/main/java/practice/CandidateValidator.java
@@ -1,5 +1,28 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MINIMUM_AGE = 35;
+    private static final String UKRAINIAN_NATIONALITY = "Ukrainian";
+    private static final int MINIMUM_YEARS_IN_UKRAINE = 10;
+    private static final String REGEX = "-";
+    private static final int POSITION_START_YEAR = 0;
+    private static final int POSITION_LAST_YEAR = 1;
+
+    private static boolean candidateYearsLiveInUkraine(Candidate candidate) {
+        String[] parts = candidate.getPeriodsInUkr().split(REGEX);
+        int fromYear = Integer.parseInt(parts[POSITION_START_YEAR]);
+        int toYear = Integer.parseInt(parts[POSITION_LAST_YEAR]);
+        return (toYear - fromYear) >= MINIMUM_YEARS_IN_UKRAINE;
+    }
+
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= MINIMUM_AGE
+                && candidate.getNationality().equals(UKRAINIAN_NATIONALITY)
+                && candidate.isAllowedToVote()
+                && candidateYearsLiveInUkraine(candidate);
+    }
 }
--- a/src/main/java/practice/StreamPractice.java
+++ b/src/main/java/practice/StreamPractice.java
@@ -1,8 +1,11 @@
 package practice;
 
-import java.util.Collections;
+import java.util.Arrays;
 import java.util.List;
+import java.util.stream.Collectors;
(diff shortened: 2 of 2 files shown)


src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +80,13 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getAge() > femaleAge
+                        && person.getCats() != null
```
olekskov, line 85: Use '>=' instead of '>' to include women with age equal to femaleAge.

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,28 @@
...
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MINIMUM_AGE = 35;
+    private static final String UKRAINIAN_NATIONALITY = "Ukrainian";
+    private static final int MINIMUM_YEARS_IN_UKRAINE = 10;
+    private static final String REGEX = "-";
```
olekskov, line 10: Consider renaming REGEX to a more descriptive name like DATE_SEPARATOR since it specifically represents the separator used in the date string.

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +32,11 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(n -> n % 2 != 0 ? numbers.get(n) - 1 : numbers.get(n))
+                .filter(n -> n % 2 != 0)
+                .average()
+                .getAsDouble();
```
olekskov, line 39: Use .orElseThrow() instead of .getAsDouble() to handle the case when there are no elements in the stream.

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +48,11 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getAge() <= toAge
+                        && person.getAge() >= fromAge
+                        && person.getSex().equals(Person.Sex.MAN))
```
olekskov, line 54: It's generally better to call .equals() on enum when we have such opportunity as we're sure it has proper value
```suggestion
                        && Person.Sex.MAN.equals(person.getSex())
```

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,28 @@
...
+    private static final int POSITION_START_YEAR = 0;
+    private static final int POSITION_LAST_YEAR = 1;
+
+    private static boolean candidateYearsLiveInUkraine(Candidate candidate) {
```
olekskov, line 14: can't see why do we need this method to be static

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +17,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(n -> Arrays.stream(n.split(",")))
```
olekskov, line 21: please recheck all occurrences of n
'number' or 'index' will be much more readable

olekskov (changes requested):

olekskov (approved):
Good job!

PR #699, opened 2023-04-11
stream practice implementations

Dmytro-Byrka:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +21,14 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        Optional<Integer> result = numbers.stream()
+                .map(n -> n.split(","))
+                .flatMap(Arrays::stream)
+                .map(Integer::parseInt)
+                .filter(n -> n % 2 == 0)
+                .min(Comparator.comparing(Integer::valueOf));
```
uliana-andrushchuk, line 29:
```suggestion
                .min();
```
You do not need to add comparator here

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +71,13 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
...
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        List<Person> people = peopleList.stream()
+                .filter(p -> (p.getSex().equals(Person.Sex.MAN)
+                        && p.getAge() >= fromAge && p.getAge() <= maleToAge)
+                        || (p.getSex().equals(Person.Sex.WOMAN)
+                        && p.getAge() >= fromAge && p.getAge() <= femaleToAge))
+                .collect(Collectors.toList());
+        return people;
```
uliana-andrushchuk, line 80: You can return the value directly, without creating variable

uliana-andrushchuk: Same for others methods

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +108,10 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(c -> new CandidateValidator().test(c))
```
uliana-andrushchuk, line 112:
```suggestion
        CandidateValidator candidateValidator = new CandidateValidator();
        return candidates.stream()
                .filter(candidateValidator)
```

uliana-andrushchuk (changes requested):

src/main/java/practice/StreamPractice.java
Dmytro-Byrka: But it requires argument, i can simplify it to : min(Integer::compare), it will be ok?

uliana-andrushchuk (approved):

PR #701, opened 2023-04-18
Hw 1 solution

BetelgeuseUA:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,35 @@
...
+
+    @Override
+    public Predicate<Candidate> negate() {
+        return null;
+    }
+
+    @Override
+    public Predicate<Candidate> or(Predicate<? super Candidate> other) {
+        return null;
+    }
```
kozhukhovsky, line 28:
```suggestion
```

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,35 @@
...
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private Predicate<Candidate> toVote = Candidate::isAllowedToVote;
+    private Predicate<Candidate> age = a -> a.getAge() >= 35;
+    private Predicate<Candidate> isNationality = c -> c.getNationality().equals("Ukrainian");
+    private Predicate<Candidate> period = p -> {
+        String[] periodOfYear = p.getPeriodsInUkr().split("-");
+        return Integer.parseInt(periodOfYear[1]) - Integer.parseInt(periodOfYear[0]) >= 10;
+    };
```
kozhukhovsky, line 13: So it is rather bad to read, because you need to look up at each condition, what is happening there. It's better to leave all this directly in the method, but put the indices, age, Ukrainian into constants

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,24 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
```
kozhukhovsky, line 22: Everything can be done much easier. Get stream numbers, then flatmap into a stream of strings broken by ",".
Then we do int and filter by condition. Everything can be done in one stream.

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +46,31 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
```
kozhukhovsky, line 48: The same thing, think about how you can do it in one stream

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +82,10 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
```
kozhukhovsky, line 84: good

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +100,12 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
```
kozhukhovsky, line 101: good

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +114,12 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
```
kozhukhovsky, line 116: good

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +135,11 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
```
kozhukhovsky, line 137: good

kozhukhovsky (changes requested):

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,24 @@ public class StreamPractice {
...
+                .mapToInt(Integer::intValue)
+                .filter(e -> e % 2 == 0)
+                .min()
+                .orElse(Integer.MIN_VALUE);
+        if (result == 0) {
+            throw new RuntimeException("Can't get min value from list: "
+                    + numbers.toString());
+        } else {
+            return result;
+        }
```
okuzan, line 39: still heavily over-engineered.  You have three operations with integers, and you ended the stream with the terminal operation collect, and then started it again. 
Let's do it simpler:
```
        return numbers.stream()
                .flatMap(e -> Arrays.stream(e.split(",")))
                .map(e -> Integer.valueOf(e))
                .filter(i -> i % 2 == 0)
                .min(Integer:: compare)
                .orElseThrow(()
```

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +45,24 @@ public int findMinEvenNumber(List<String> numbers) {
...
+                    return e;
+                }).collect(Collectors.toList()).stream()
+                .filter(e -> e % 2 == 1)
+                .collect(Collectors.toList());
+        if (result.isEmpty()) {
+            throw new NoSuchElementException("No such element");
+        }
+        return result.stream()
+                .mapToDouble(Integer::doubleValue)
+                .sum() / result.size();
```
okuzan, line 65: you call stream here 3 times. Try to rework this given the solution from last task. Hint: here it's enough to use IntStream.range(), map(), filter(), average(), getAsDouble()

okuzan (changes requested):

olekskov (approved):
Good job, approved.
Further please try to avoid 1-letter variables even in streams
'number' is better than 'n', 'person' is better than 'p', etc
It's easy to read when code has been just written but in couple of weeks you won't remember what the hell is 'e' :)

PR #706, opened 2023-04-26
added solution for StreamPractice and Impl

yuriy-kononchuk:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,25 @@
...
+                && candidate.getNationality().equals("Ukrainian")
+                && getLivingYearsInUkraine(candidate.getPeriodsInUkr()) >= MIN_LIVING_PERIOD;
+    }
+
+    private static int getLivingYearsInUkraine(String period) {
+        String[] livingYears = period.split("-");
+        return Integer.parseInt(Arrays.stream(livingYears)
+                .reduce((year1, year2) -> String.valueOf(Integer.parseInt(year2)
+                        - Integer.parseInt(year1)))
+                .get());
```
uliana-andrushchuk, line 23:
```suggestion
        return Integer.parseInt(livingYears[1]) - Integer.parseInt(livingYears[0])
```
It can be simplified

PR #708, opened 2023-05-01
hw-jv-stream

VolodymyrDavydov:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,22 @@
...
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MIN_AGE = 35;
+    private static final int LIVE_IN_UKRAINE = 10;
+    private static final String NATIONALITY = "Ukrainian";
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] periodInUkr = candidate.getPeriodsInUkr().split("-");
+        int startPeriodInUkr = Integer.parseInt(periodInUkr[0]);
+        int finishPeriodInUkr = Integer.parseInt(periodInUkr[1]);
```
kozhukhovsky, line 15: It would be nice to put the indexes into constants

kozhukhovsky (approved):

PR #715, opened 2023-05-05
Solution

OlegMR0:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,78 @@
...
+                .map(Candidate::getName)
+                .sorted()
+                .collect(Collectors.toList());
+    }
+
+    private List<Person> selectPersonByAgeAndSex(List<Person> list,
+                                                 int fromAge, int toAge, Person.Sex sex) {
+        return list.stream()
+                .filter(p -> p.getSex().equals(sex) && p.getAge() <= toAge && p.getAge() >= fromAge)
+                .collect(Collectors.toList());
```
okuzan, line 76: no need to extract it to a separate method

okuzan (approved):
Good job!

PR #716, opened 2023-05-06
hw-17-solution

vrnvv:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,78 @@
...
-    /**
-     * Given list of strings where each element contains 1+ numbers:
-     * input = {"5,30,100", "0,22,7", ...}
-     * return min integer value. One more thing - we're interested in even numbers.
-     * If there is no needed data throw RuntimeException with message
-     * "Can't get min value from list: < Here is our input 'numbers' >"
-     */
+
+    private static final String ERR_MESSAGE = "Can't get min value from list: ";
+    private static int CURRENT_INDEX;
```
okuzan, line 14: remove this variable

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,78 @@
...
     }
 
-    /**
-     * Given a List of Integer numbers,
-     * return the average of all odd numbers from the list or throw NoSuchElementException.
-     * But before that subtract 1 from each element on an odd position (having the odd index).
-     */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        CURRENT_INDEX = 0;
```
okuzan, line 26: remove this variable

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,78 @@
...
+        return numbers.stream()
+                .mapToInt(n -> (current_index(numbers.size()) % 2 == 0) ? n - 1 : n)
```
okuzan, line 28: Replace the 'current_index()' function with IntStream.range() to avoid using static variable.

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,78 @@
...
-        return Collections.emptyList();
+        CandidateValidator candidateValidator = new CandidateValidator();
+        return candidates.stream()
+                .filter(candidateValidator)
+                .map(Candidate::getName)
+                .sorted()
+                .collect(Collectors.toList());
+    }
+
+    private int current_index(int size) {
```
okuzan, line 72: Remove the 'current_index()' function since it's not the best practice for handling indexes in streams.

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,25 @@
...
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= CORRECT_AGE
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals(CORRECT_NATIONALITY)
+                && calculatePeriodInUkraine(candidate.getPeriodsInUkr()
+                .split("-"));
+    }
+
+    private boolean calculatePeriodInUkraine(String[] periodsInUkr) {
```
okuzan, line 20: this method name implies that you would return integer :)

src/test/java/core/basesyntax/StreamPracticeTest.java
```diff
@@ -96,6 +96,7 @@ public void getOddNumsAverage_basicData() {
     public void getOddNumsAverage_twoEven() {
         Double expected = 4.0;
         List<Integer> digits = Arrays.asList(7, 2, 3, 4, 5, 6);
+
```
okuzan, line 99: make sure to not make unintended changes. Review what filed are modified when committing

src/main/java/practice/StreamPractice.java
okuzan: also, we use cameCase for method names, not snake_case

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,25 @@
...
+    private static final String CORRECT_NATIONALITY = "Ukrainian";
+    private static final int MIN_LIVING_AGE_IN_UKR = 10;
+
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= CORRECT_AGE
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals(CORRECT_NATIONALITY)
+                && calculatePeriodInUkraine(candidate.getPeriodsInUkr()
+                .split("-"));
```
okuzan, line 17: make your helper method accept one string and split it there
- should also be extracted to DELIMITER constant

okuzan (changes requested):

Ivan95kos (approved):
Great job

PR #718, opened 2023-05-11
created the CandidateValidator class, create a test method that validates the candidate, as well as two constants min age and lived in Ukraine  solved the task according to the conditions

TemkaHillfinger:


Ivan95kos (approved):
Great job

PR #721, opened 2023-05-13
Hw jv stream GitHub practice

urlo0k:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,34 @@
...
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import java.util.stream.Stream;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int AGE_MIN = 35;
+    private static final int ZERO = 0;
+    private static final int ONE = 1;
```
okuzan, line 10: naming 0 a ZERO doesn't add more context :) use constant naming as an opportunity to describe its purpose, here it's a start year index

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(str -> Arrays.stream(str.split(",")))
+                .map(Integer::parseInt)
+                .filter(a -> a % 2 == 0)
```
okuzan, line 24: please rename short variables' names in the code to something more readable, like integer, number

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +33,11 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(i -> i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i))
+                .filter(v -> v % 2 != 0)
+                .average()
+                .getAsDouble();
```
okuzan, line 40: Use .orElseThrow() instead of .getAsDouble() to handle the case when there are no elements in the stream.

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +49,19 @@ public Double getOddNumsAverage(List<Integer> numbers) {
...
-        return Collections.emptyList();
+        class PersonPredicate implements Predicate<Person> {
+            @Override
+            public boolean test(Person person) {
+                return person.getSex() == Person.Sex.MAN
+                        && person.getAge() >= fromAge
+                        && person.getAge() <= toAge;
+            }
+        }
+
```
okuzan, line 60: just embed this in filter

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +76,25 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
...
+                            && person.getAge() >= fromAge
+                            && person.getAge() <= femaleToAge;
+                } else {
+                    return person.getSex() == Person.Sex.MAN
+                            && person.getAge() >= fromAge
+                            && person.getAge() <= maleToAge;
+                }
+            }
+        }
+
```
okuzan, line 93: move the condition to filter, having a local class is less readable

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +122,11 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        //Predicate<Candidate> candidateValidator = new CandidateValidator();
```
okuzan, line 125: remove the comments before committing

okuzan (changes requested):

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(string -> Arrays.stream(string.split(",")))
+                .map(Integer::parseInt)
+                .filter(number -> number % 2 == 0)
+                .reduce(Integer::min)
```
kozhukhovsky, line 25:
```suggestion
                .min()
```

kozhukhovsky (approved):

kozhukhovsky (approved):

PR #728, opened 2023-05-21
task completed

marimarig:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,68 @@
...
-     * select from the List only men whose age is from `fromAge` to `toAge` inclusively.
-     * <p>
-     * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
-     */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getAge() >= fromAge
+                        && p.getAge() <= toAge
+                        && p.getSex() == Person.Sex.MAN)
```
sZlishchev, line 38: Create predicate for these pourose it can improve readability and also check the same logic in following methods

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,68 @@
...
+
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(n -> n.split(NUMBERS_SEPARATOR))
+                .flatMap(Arrays::stream)
+                .mapToInt(Integer::valueOf)
+                .filter(s -> s % 2 == 0)
+                .min()
+                .orElseThrow(() -> new RuntimeException("Can't get min value from list: "
```
sZlishchev, line 22: In my opinion it would look better if all the method started from the same place as .stream(), just add a few tabs to each

sZlishchev (changes requested):
Good job! Let's slightly improve your solution

sZlishchev (approved):
Great job! Keep pushing)))

maxline (changes requested):
Good job! Left just a minor comments. Let's improve the solution

sarakhmen (approved):
Good job ;)

PR #730, opened 2023-05-22
Implemented methods for Stream API practice;

viprko:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,25 @@
...
+        return candidate.getAge() >= MIN_AGE
+                && candidate.getNationality().equals(PROPER_NATIONALITY)
+                && candidate.isAllowedToVote()
+                && calculatePeriodInUkr(candidate.getPeriodsInUkr())
+                >= MIN_LIVING_PERIOD_IN_UKRAINE;
+    }
+
+    private int calculatePeriodInUkr(String period) {
+        String[] firstAndLastYear = period.split(DELIMITER);
+        return Integer.parseInt(firstAndLastYear[1]) - Integer.parseInt(firstAndLastYear[0]);
```
marimarig, line 23: Numbers 1 and 0 - also Magic numbers, make them private static final

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,25 @@
...
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    public static final String PROPER_NATIONALITY = "Ukrainian";
+    public static final String DELIMITER = "-";
+    public static final int MIN_AGE = 35;
+    public static final int MIN_LIVING_PERIOD_IN_UKRAINE = 10;
```
marimarig, line 10: maybe better to do private

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +21,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(s -> Arrays.stream(s.split(",")))
```
marimarig, line 25: "," - make also private static final

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +36,12 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(i -> i % 2 == 1 ? numbers.get(i) - 1 : numbers.get(i))
+                .filter(n -> n % 2 == 1)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
marimarig, line 43: .orElseThrow replacement for .getAsDouble();

marimarig (changes requested):
Good, but you can improve your solution

src/main/java/practice/StreamPractice.java
viprko: Sorry, but the task has a condition about throwing an exception.

marimarig: [https://mate-academy.github.io/jv-program-common-mistakes/java-core/java-eight-part-two/stream-api](url) 3 line

viprko: Ok. But I strongly disagree with this common mistakes part. 1st. The volume of code is the same. 2nd. Readability is worse because when users read this method, it's clear that it can throw an exception, and users shouldn't remember what getAsDouble or getAsInt or etc. do. 3rd. I can use lambda and change the message of exception, instead of throws "No value present". Of course, maybe I haven't enough knowledge about advantages of using getAs..., but for now it's my opinion.

marimarig: Ok, I respect your opinion, leave this part for mentors

marimarig (approved):
Good

src/main/java/practice/StreamPractice.java
maxline: 2nd. Java libraries (not only for Stream but in general) have lot's of methods that can throw runtime exceptions, this doesn't mean that developer should try avoid to use them. The question is how this exceptions will be processed later.
3d. If you need change the exception message surely you can use `orElseThrow()`, in other case it is enough to use default implementation that present in getAsDouble(). The common approach is try to use maximum of methods that already implemented in libraries when it possible and to not implement manually already existing logic.

PR #731, opened 2023-05-23
implemented task

khrystyna-dev:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,25 @@
...
+                && candidate.getNationality().equals(NATIONALITY)
+                && livingYear(candidate) > MIN_LIVING_YEAR;
+    }
+
+    private int livingYear(Candidate candidate) {
+        return Integer.parseInt(candidate.getPeriodsInUkr()
+                        .substring(candidate.getPeriodsInUkr().indexOf('-') + 1))
+                - Integer.parseInt(candidate.getPeriodsInUkr()
+                        .substring(0, candidate.getPeriodsInUkr().indexOf('-')));
+    }
```
Pkrasota, line 24: Instead of using substrings and indexes, you can use split

`    String[] years = candidate.getPeriodsInUkr().split("-"); `
`    int startYear = Integer.parseInt(years[0]);`
`    int endYear = Integer.parseInt(years[1]);`
`    return endYear - startYear;`

The code will be more simpler

Pkrasota (changes requested):
Great job :)
Just some recommendation

src/main/java/practice/CandidateValidator.java
chrisnyrka: Your version has good code readability, but my version uses less memory because I don't create an redundant array. But I don't know how better, let the mentor judge us))

Pkrasota (approved):
Great job :)

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,78 @@
...
-     * return the average of all odd numbers from the list or throw NoSuchElementException.
-     * But before that subtract 1 from each element on an odd position (having the odd index).
-     */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(i -> i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i))
+                .filter(e -> e % 2 != 0)
+                .average()
+                .getAsDouble();
```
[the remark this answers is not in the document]
rovein: `getAsDouble` will throw an exception if the value is not present.

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,25 @@
...
+
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= MIN_AGE
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals(NATIONALITY)
+                && livingYear(candidate) > MIN_LIVING_YEAR;
+    }
+
+    private int livingYear(Candidate candidate) {
```
rovein, line 19:
```suggestion
    private int getLivingYear(Candidate candidate) {
```

src/main/java/practice/CandidateValidator.java
rovein: I agree with Pasha. Let's better introduce variables, the code will be much more readable.

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,77 @@
...
+    private final CandidateValidator candidateValidator;
+
+    public StreamPractice() {
+        this.candidateValidator = new CandidateValidator();
+    }
+
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(e -> Arrays.stream(e.split(",")))
```
rovein, line 23: Hardcoded constant `","`.

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,77 @@
...
-     * But before that subtract 1 from each element on an odd position (having the odd index).
-     */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(i -> i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i))
+                .filter(e -> e % 2 != 0)
+                .average()
+                .orElseThrow(
+                        () -> new NoSuchElementException("No odd numbers found in the list."));
```
rovein, line 37: If you need to throw exception when no result found after stream execution sometimes get()/getAsDouble() may help, try to google what does it do.

So, let's better use `getAsDouble()`.

rovein (approved):
Great job! Pay attention to some minor comments.

PR #744, opened 2023-05-25
done stream practice

ViktoriiaSemenchenko:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,11 +1,16 @@
...
 import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
 import model.Candidate;
+import model.Cat;
 import model.Person;
 
 public class StreamPractice {
+    private static final String SEPARATOR = ",";
```
sZlishchev, line 12: I think COMA_SEPARATOR would be better

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +50,11 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex() == Person.Sex.MAN
+                        && p.getAge() >= fromAge
+                        && p.getAge() <= toAge)
```
sZlishchev, line 56: Create Predicate for this

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +69,9 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> isAgeInRange(person, fromAge, femaleToAge, maleToAge))
```
sZlishchev, line 73: It'd be easier to create Predicate instead of creating a new method. Check it in all the similar methods

sZlishchev (changes requested):

sZlishchev (approved):
Excellent job! Keep pushing))

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
```
sarakhmen, line 22: let's remove this redundant comments :)

sarakhmen (approved):
Overall good job, you can remove redundant comment if you want ;)

PR #746, opened 2023-05-25
Create StreamPractice

DV7Art:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +17,14 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        final String Separator = ",";
```
ArinaPolna, line 20: you can make this variable a constant(private static final)

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +67,12 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> (p.getSex() == Person.Sex.MAN
+                        && p.getAge() >= fromAge & p.getAge() <= maleToAge)
+                        || (p.getSex() == Person.Sex.WOMAN
+                        && p.getAge() >= fromAge & p.getAge() <= femaleToAge))
+                .toList();
```
ArinaPolna, line 75: to my mind better to use collect(Collectors.toList()) to prevent errors which may be caused when you will change your list

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +81,11 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex() == Person.Sex.WOMAN && p.getAge() >= femaleAge)
+                .flatMap(n -> n.getCats().stream())
+                .map(Cat::getName)
+                .toList();
```
ArinaPolna, line 88: the same

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +49,10 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex() == Person.Sex.MAN
+                        && p.getAge() >= fromAge & p.getAge() <= toAge)
+                .toList();
```
ArinaPolna, line 55: see comment on line 75

ArinaPolna (changes requested):

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,27 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.Arrays;
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+
```
rovein, line 8: Remember about empty lines. Let's remove them.

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,27 @@
...
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= MIN_AGE
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals(NATIONALITY)
+                && isValidPeriodInUkr(candidate) >= VALID_PERIOD;
+    }
+
+    public int isValidPeriodInUkr(Candidate candidate) {
```
rovein, line 21:
```suggestion
    public int getPeriodInUkr(Candidate candidate) {
```

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +35,11 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .mapToDouble(i -> (i % 2 != 0) ? numbers.get(i) - 1 : numbers.get(i))
+                .filter(number -> number % 2 != 0)
+                .average()
+                .orElseThrow(() -> new NoSuchElementException("No odd numbers found in the list"));
```
rovein, line 42: If you need to throw exception when no result found after stream execution sometimes get()/getAsDouble() may help, try to google what does it do.

So, let's better use `getAsDouble()`.

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +69,12 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> (p.getSex() == Person.Sex.MAN
+                        && p.getAge() >= fromAge & p.getAge() <= maleToAge)
```
rovein, line 74: Let's better use logical "and" rather than bitwise one.
```suggestion
                .filter(p -> (p.getSex() == Person.Sex.MAN
                        && p.getAge() >= fromAge && p.getAge() <= maleToAge)
```

rovein (changes requested):
Good job! But let's improve your solution ;)

src/main/java/practice/StreamPractice.java
DV7Art: An alternate method for replacing the Optional.get (which more likely than not fails the user's intentions with a NoSuchElementException) is with a more verbose API introduced in JDK10 termed as Optional.orElseThrow()
Optional.get() is an "attractive nuisance" and is too tempting for programmers, leading to frequent errors.

rovein: Yes, but we are talking about the task in terms of a checklist :) In such case we need to use `getAsDouble` and under the hood it will throw the exception by you.

rovein (approved):
Great job, well done)

PR #752, opened 2023-05-26
StreamPractice task is done

m-pnsk:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
```
nechupara, line 21: remove all the unnecessary comments

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,13 @@ public class StreamPractice {
...
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(n -> Arrays.stream(n.split(SEPARATOR)))
+                .mapToInt(Integer::parseInt)
+                .filter(i -> i % 2 == 0)
+                .min()
+                .orElseThrow(() ->
+                        new RuntimeException("Can't get min value from list: " + numbers));
```
nechupara, line 29: exception message doesn't look  like in the task
```suggestion
                        new RuntimeException("Can't get min value from list: < Here is our input " + numbers + " >"));
```

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +35,10 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(n -> n % 2 == 0 ? numbers.get(n) : numbers.get(n) - 1)
+                .filter(n -> n % 2 != 0)
+                .average().orElseThrow(NoSuchElementException::new);
```
nechupara, line 41: first of all, I think, you can move last expression (.orElseThrow) to the next line.
And secondly I think that you can simplify it by replacing .orElseThrow to just .getAsDouble(), Because .getAsDouble() will throw NoSuchElementException in case empty number
```suggestion
                .average()
                .getAsDouble();
```

nechupara (changes requested):
good job! consider minor improvments

src/main/java/practice/StreamPractice.java
m-pnsk: image
I don't agree. Look at common misstakes

nechupara (approved):
nice )

rovein (approved):
Great job, see one minor comment.

PR #755, opened 2023-05-26
Completed the task StreamPractice

VRuban373:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,64 @@
...
-    /**
-     * Given a List of Integer numbers,
-     * return the average of all odd numbers from the list or throw NoSuchElementException.
-     * But before that subtract 1 from each element on an odd position (having the odd index).
-     */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .mapToDouble(i -> (i % 2 == 1) ? numbers.get(i) - 1 : numbers.get(i))
+                .filter(n -> n % 2 == 1)
```
nechupara, line 25: there is no need to use `mapToDouble`, because `.average() `method returns `OptionDouble` anyway. 
Also I suggest you to convert logic to `i % 2 == 0` for better code understanding, but it's optional, it's up to you

```suggestion
                .map(i -> (i % 2 == 0 ? numbers.get(i) : numbers.get(i) - 1))
                .filter(n -> n % 2 != 0)
```

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,64 @@
...
-     * (from 18 y.o. and to 60 y.o. for men and to 55 y.o. for women inclusively).
-     */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getSex() == Person.Sex.MAN && person.getAge() >= fromAge
+                        && person.getAge() <= maleToAge
+                        || person.getSex() == Person.Sex.WOMAN && person.getAge() >= fromAge
+                        && person.getAge() <= femaleToAge)
```
nechupara, line 43: this is optional, it's up to you: Both women and men should be older than `fromAge` so there is no need to check it twice.
Let's take a look on such an example:
```suggestion
                .filter(i -> i.getAge() >= fromAge
                        && (i.getSex() == Person.Sex.MAN && i.getAge() <= maleToAge
                        || i.getSex() == Person.Sex.WOMAN && i.getAge() <= femaleToAge)
                )
```

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,64 @@
...
-     * and live in Ukraine for 10 years. For the last requirement use field periodsInUkr,
-     * which has following view: "2002-2015"
-     * We want to reuse our validation in future, so let's write our own impl of Predicate
-     * parametrized with Candidate in CandidateValidator.
-     */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        CandidateValidator validator = new CandidateValidator();
+        return candidates.stream()
+                .filter(validator)
```
nechupara, line 59: this is also up to you. But in my opinion there is no need to create separate variable `validator`, you could create validator object directly in `filter` method call
```suggestion
                return candidates.stream()
                .filter(new CandidateValidator())
```

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,28 @@
...
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MINIMUM_AGE = 35;
+    private static final String NATIONALITY = "Ukrainian";
+    private static final int MINIMUM_LIVE_IN_UA = 10;
+    private static final int INDEX_FROM_YEAR = 0;
+    private static final int INDEX_TO_YEAR = 1;
+    private static final String HYPHEN = "-";
```
nechupara, line 12: I suggest you to rename this variable into more abstract term like `YEAR_SEPARATOR` or something like this

nechupara (changes requested):
good job! but let's improve your solution a little bit )

nechupara (changes requested):
try to reject

nechupara (approved):
Nice!

maxline (approved):
Well done! Approved

PR #759, opened 2023-06-02
StreamPractice

RostyslavOnysh:


src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +72,13 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
```
overpathz, line 75: Personally, I think it's a little hard to read.
Consider smth like this(or improve):

return peopleList.stream()
                .filter(person -> person.getAge() >= fromAge)
                .filter(person -> person.getSex().equals(Person.Sex.MAN)
                ? person.getAge() <= maleToAge : person.getAge() <= femaleToAge)

overpathz (approved):
Greate job. Just one minor comment.

PR #763, opened 2023-06-05
work with stream done.

Hlib13:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,38 @@
...
+            int startYear = Integer.parseInt(years[0]);
+            int endYear = Integer.parseInt(years[1]);
+
+            Year startDate = Year.of(startYear);
+            Year endDate = Year.of(endYear);
+
+            totalYearsInUkraine = endDate.compareTo(startDate);
+        }
+
+        return totalYearsInUkraine >= 10;
```
okuzan, line 36: integers are enough to compare directly, without Year

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,17 @@ public class StreamPractice {
...
+                .flatMap(numberStr -> Arrays.stream(numberStr.split(",")))
+                .map(String::trim)
+                .map(Integer::parseInt)
+                .filter(num -> num % 2 == 0)
+                .min(Integer::compare);
+
+        if (minEven.isEmpty()) {
+            throw new RuntimeException("Can't get min value from list: " + numbers);
+        }
+        return minEven.get();
```
okuzan, line 32: use orElseThrow

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +81,14 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
...
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> (person.getAge() >= fromAge
+                        && person.getSex() == Person.Sex.MAN
+                        && person.getAge() <= maleToAge)
+                        || (person.getAge() >= fromAge
+                        && person.getSex() == Person.Sex.WOMAN
+                        && person.getAge() <= femaleToAge))
```
okuzan, line 90: from age is common for both sexes

okuzan (changes requested):

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,16 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        Optional<Integer> minEven = Optional.ofNullable(numbers.stream()
+                .flatMap(numberStr -> Arrays.stream(numberStr.split(",")))
+                .map(String::trim)
+                .map(Integer::parseInt)
```
aaananas, line 24: combine those map() methods

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +36,13 @@ public int findMinEvenNumber(List<String> numbers) {
...
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        double count = IntStream.range(0, numbers.size())
+                .map(n -> (n % 2 == 0) ? numbers.get(n) : numbers.get(n) - 1)
+                .filter(n -> n % 2 == 1)
+                .average()
+                .getAsDouble();
+
+        return count;
```
aaananas, line 45:
```suggestion
        return IntStream.range(0, numbers.size())
                .map(n -> (n % 2 == 0) ? numbers.get(n) : numbers.get(n) - 1)
                .filter(n -> n % 2 == 1)
                .average()
                .getAsDouble();
```

aaananas (changes requested):

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,15 @@ public class StreamPractice {
...
-        return 0;
+        Optional<Integer> minEven = Optional.ofNullable(numbers.stream()
+                .flatMap(numberStr -> Arrays.stream(numberStr.split(",")))
+                .map(numberStr -> Integer.parseInt(numberStr.trim()))
+                .filter(num -> num % 2 == 0)
+                .min(Integer::compare)
+                .orElseThrow(() -> new RuntimeException("Can't get min value from list: "
+                        + numbers)));
+
+        return minEven.get();
```
overpathz, line 29: You don't need this line. Return straight like it is.
`return numbers.stream()...<other_statements>....min().orElseThrow(() -> new RuntimeException("Can't get min value from list: " + numbers));`

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +35,11 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
```
overpathz, line 37: What if we don't find our value? Should you throw some kind of exception?

olekskov (approved):
Good job

PR #769, opened 2023-06-16
Ready for check

AnatolyNz:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,32 @@
...
+    private static final int ONE = 1;
+    private static final int TERM_LIVE_IN_UKRAINE = 10;
+    private static final String NATIONALITY_CANDIDATE = "Ukrainian";
+
+    @Override
+    public boolean test(Candidate candidate) {
+        boolean period = Stream.of(candidate)
+                .anyMatch(person -> {
+                    String periodsInUkr = person.getPeriodsInUkr();
+                    if (periodsInUkr != null) {
```
okuzan, line 19: you don't need that if

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,32 @@
...
+                        String[] years = periodsInUkr.split("-");
+                        int startYear = Integer.parseInt(years[NULL]);
+                        int endYear = Integer.parseInt(years[ONE]);
+                        return ((endYear - startYear) >= TERM_LIVE_IN_UKRAINE);
```
okuzan, line 23: it's hardly readable. Better extract to a private utility method. Null and zero are very different things :). Better give more informative names , like START_INDEX, END_INDEX

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +22,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(n -> n.split(","))
+                .flatMap(n -> Arrays.stream(n))
+                .mapToInt(Integer::valueOf)
+                .filter(n -> n % 2 == 0)
```
okuzan, line 29: fix everywhere - don't name variables in stream chains with one letter - make it `number`, `person` etc

src/main/java/practice/StreamPractice.java
```diff
@@ -50,31 +71,47 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
...
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex() == Person.Sex.MAN && p.getAge() >= fromAge
+                        && p.getAge() <= maleToAge
+                        && p.getAge() >= START_WORK_AGE && p.getAge() <= AGE_MAX_WORK_MEN
+                        || p.getSex() == Person.Sex.WOMAN && p.getAge() >= fromAge
+                        && p.getAge() <= femaleToAge
+                        && p.getAge() >= START_WORK_AGE && p.getAge() <= AGE_MAX_WORK_WOMEN)
```
okuzan, line 80: extract common condition from both clauses

src/main/java/practice/StreamPractice.java
```diff
@@ -23,25 +37,32 @@ public int findMinEvenNumber(List<String> numbers) {
...
+                  .collect(Collectors.toList());
     }
 
     /**
-     * Given a List of `Person` instances (having `name`, `age` and `sex` fields),
-     * for example, `Arrays.asList( new Person(«Victor», 16, Sex.MAN),
-     * new Person(«Helen», 42, Sex.WOMAN))`,
+     * Given a List of `model.Person` instances (having `name`, `age` and `sex` fields),
+     * for example, `Arrays.asList( new model.Person(«Victor», 16, Sex.MAN),
+     * new model.Person(«Helen», 42, Sex.WOMAN))`,
```
okuzan, line 65: be careful when using IDEA's renaming :)

src/main/java/model/Person.java
```diff
@@ -62,7 +62,7 @@ public int hashCode() {
 
     @Override
     public String toString() {
-        return "Person{"
+        return "model.Person{"
```
okuzan, line 65: not intended change

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,33 @@
...
+    private static final int TERM_LIVE_IN_UKRAINE = 10;
+    private static final String NATIONALITY_CANDIDATE = "Ukrainian";
+
+    @Override
+    public boolean test(Candidate candidate) {
+        boolean period = Stream.of(candidate)
+                .anyMatch(person -> {
+                    String periodsInUkr = person.getPeriodsInUkr();
+                    return checkPeriodsInUkraine(periodsInUkr);
+                });
```
okuzan, line 20: you don't need this, use utility method in return statement directly

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +22,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(number -> number.split(","))
```
okuzan, line 26: make string constant

src/main/java/practice/StreamPractice.java
```diff
@@ -23,25 +37,33 @@ public int findMinEvenNumber(List<String> numbers) {
...
+                .orElseThrow(() -> new NoSuchElementException("Array is empty"));
     }
 
     /**
-     * Given a List of `Person` instances (having `name`, `age` and `sex` fields),
-     * for example, `Arrays.asList( new Person(«Victor», 16, Sex.MAN),
-     * new Person(«Helen», 42, Sex.WOMAN))`,
+     * Given a List of `model.Person` instances (having `name`, `age` and `sex` fields),
+     * for example, `Arrays.asList( new model.Person(«Victor», 16, Sex.MAN),
+     * new model.Person(«Helen», 42, Sex.WOMAN))`,
```
okuzan, line 51: try to retract these changes

src/main/java/practice/StreamPractice.java
```diff
@@ -50,31 +72,50 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getAge() >= START_WORK_AGE
```
okuzan, line 76: extract from age, it's common for both sexes

okuzan (changes requested):

src/main/java/practice/StreamPractice.java
AnatolyNz: I don't understand what should do me

okuzan (approved):
Good job

PR #770, opened 2023-06-17
Hometask completed jv-stream

MarynaKushniruk:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,15 @@ public class StreamPractice {
...
-        return 0;
+
+        return numbers.stream()
+                .map(s -> s.split(","))
+                .flatMap(Arrays::stream)
+                .mapToInt(Integer::parseInt)
+                .filter(i -> i % 2 == 0)
+                .min()
+                .orElseThrow(() -> new RuntimeException("Can't get min value from list: "
+                        + "< Here is our input 'numbers' >"));
```
Ivan95kos, line 31: All magic numbers, strings, symbols in your code should be constants.
do not use the abbreviation names, fix everywhere
```suggestion
        return numbers.stream()
                .map(line -> line.split(SEPARATOR))
                .flatMap(Arrays::stream)
                .mapToInt(Integer::parseInt)
                .filter(number -> number % 2 == 0)
                .min()
                .orElseThrow(() -> new RuntimeException("Can't get min value from list: "
                        + numbers));
```

Ivan95kos (changes requested):

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,19 @@
...
+    private static final int MIN_AGE_FOR_CANDIDATE = 35;
+    private static final int MIN_YEAR_LIVE_IN_UKRAINE = 10;
+
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= MIN_AGE_FOR_CANDIDATE && candidate.isAllowedToVote()
+                && candidate.getNationality().equals("Ukrainian")
+                && (Integer.parseInt(candidate.getPeriodsInUkr().split("-")[1])
+                - (Integer.parseInt(candidate.getPeriodsInUkr()
+                .split("-")[0])) > MIN_YEAR_LIVE_IN_UKRAINE);
```
okuzan, line 16: indexes, delimiter should be constants too

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +107,12 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
...
-        return Collections.emptyList();
+
+        return candidates.stream().filter(candidate -> new CandidateValidator().test(candidate))
+                .map(Candidate::getName)
+                .sorted(Comparator.naturalOrder())
+                .collect(Collectors.toList());
     }
 }
+
+
```
okuzan, line 118: remove

okuzan (approved):
Well done!

PR #771, opened 2023-06-21
Hw solution

dmytro-voronin:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,109 @@
...
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        List<String> validCandidates = new ArrayList<>();
+
+        Predicate<Candidate> eligibilityPredicate = candidate ->
+                candidate.getAge() >= 35
+                        && candidate.isAllowedToVote()
+                        && candidate.getNationality().equals("Ukrainian")
+                        && isPeriodInUkraineLongerThanTenYears(candidate.getPeriodsInUkr());
+        for (Candidate candidate : candidates) {
```
okuzan, line 86: image

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,37 @@
...
+        if (periodsInUkr != null && !periodsInUkr.isEmpty()) {
+            String[] years = periodsInUkr.split("-");
+            if (years.length == 2) {
+                int startYear = Integer.parseInt(years[0]);
+                int endYear = Integer.parseInt(years[1]);
+                int residencyYears = now.getYear() - startYear + 1;
+
+                return residencyYears >= MINIMUM_RESIDENCY_YEARS || now.getYear() <= endYear;
+            }
+        }
```
okuzan, line 34: image
when you need to assess boolean variables to return a boolean, like in your case, never use true/false, return condition directly
For example, instead of 
`if(check) return true; else return false;`
just write: `return check; `

okuzan (changes requested):

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,101 @@
...
+            int startYear = Integer.parseInt(years[0]);
+            int endYear = Integer.parseInt(years[1]);
+            int yearDifference = endYear - startYear;
+
+            return yearDifference >= 10;
+        } else {
+            return false;
+        }
     }
 }
```
okuzan, line 101: here you should just filter using candidate validator. all that logic should be there

okuzan (changes requested):
Previous comments not fixed - don't use loops in this task, each integer and string in candidate validator should be constant

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,71 @@
...
-     * Given list of strings where each element contains 1+ numbers:
-     * input = {"5,30,100", "0,22,7", ...}
-     * return min integer value. One more thing - we're interested in even numbers.
-     * If there is no needed data throw RuntimeException with message
-     * "Can't get min value from list: < Here is our input 'numbers' >"
-     */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(s -> Arrays.stream(s.split(",")))
```
Ivan95kos, line 15: All magic numbers, strings, symbols in your code should be constants.
do not use abbreviated names
```suggestion
                .flatMap(line -> Arrays.stream(line.split(SEPARATOR)))
```

Ivan95kos (approved):

PR #774, opened 2023-06-27
jv-stream-practice Solution

ArtEg0:
Added logic to CandidateValidator.
Added logic to all methods at StreamPractice.

### DIFF
--- a/src/main/java/practice/StreamPractice.java
+++ b/src/main/java/practice/StreamPractice.java
@@ -1,8 +1,12 @@
 package practice;
 
-import java.util.Collections;
+import java.util.Arrays;
 import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(s -> Arrays.stream(s.split(",")))
```
Ivan95kos, line 22: For the future, do not use abbreviation names
```suggestion
                .flatMap(line -> Arrays.stream(line.split(SEPARATOR)))
```

Ivan95kos (approved):

PR #776, opened 2023-06-29
implemented all methods by stream API

pgolibroda:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(s -> Arrays.stream(s.split(",")))
+                .mapToInt(Integer::parseInt)
+                .filter(n -> n % 2 == 0)
```
okuzan, line 24: don't use naming such as `i`, `n`, `s`, it decreases readability

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,26 @@
...
+    private static final int NEEDED_AGE = 35;
+    private static final String NEEDED_NATIONALITY = "Ukrainian";
+    private static final int NEEDED_YEARS_OF_LIVING = 10;
+    private static final String DATA_DELIMITER = "-";
+    private static final int TO_YEARS_INDEX = 1;
+    private static final int FROM_YEARS_INDEX = 0;
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] periodsToArray = candidate.getPeriodsInUkr().split(DATA_DELIMITER);
```
okuzan, line 16:
```suggestion
        String[] periodsArray = candidate.getPeriodsInUkr().split(DATA_DELIMITER);
```

okuzan (changes requested):
Good job! Please change variables naming

olekskov (approved):
Good job!

PR #777, opened 2023-06-29
added implementation CandidateValidator

krismva:
implemented stream.api

### DIFF
--- a/src/main/java/practice/CandidateValidator.java
+++ b/src/main/java/practice/CandidateValidator.java
@@ -1,5 +1,23 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MIN_AGE = 35;
+    private static final int MIN_YEARS_IN_UKRAINE = 10;
+    private static final String NATIONALITY = "Ukrainian";
+
+    @Override
+    public boolean test(Candidate candidate) {
+        int age = candidate.getAge();
+        boolean allowedToVote = candidate.isAllowedToVote();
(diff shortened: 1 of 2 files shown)


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,23 @@
...
+    private static final int MIN_AGE = 35;
+    private static final int MIN_YEARS_IN_UKRAINE = 10;
+    private static final String NATIONALITY = "Ukrainian";
+
+    @Override
+    public boolean test(Candidate candidate) {
+        int age = candidate.getAge();
+        boolean allowedToVote = candidate.isAllowedToVote();
+        String[] periods = candidate.getPeriodsInUkr().split("-");
+        int yearsInUkraine = Integer.parseInt(periods[1]) - Integer.parseInt(periods[0]);
```
okuzan, line 16: 1,0,- should be constants

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,70 @@
...
-        return 0;
+    public static void main(String[] args) {
+        System.out.println(findMinEvenNumber(List.of("1,2,3", "3,4,5")));
+    }
+
+    public static int findMinEvenNumber(List<String> numbers) {
+        return numbers.stream()
+                .flatMap(n -> Arrays.stream(n.split(",")))
+                .map(Integer::valueOf)
+                .filter(n -> n % 2 == 0)
```
okuzan, line 22: don't name variables in stream calls with one letter, make it `person`, `integer`, `number` etc

okuzan (changes requested):

olekskov (approved):
Good job!

PR #783, opened 2023-07-08
Stream practice solution

knyazkyylubomir:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +16,15 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(s -> s.split(","))
```
Ivan95kos, line 20: All magic numbers, strings, symbols in your code should be constants.
don't use short name, fix everywhere
```suggestion
                .map(line -> line.split(SEPARATOR))
```

Ivan95kos (changes requested):

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +35,11 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(index -> index % 2 == 1 ? numbers.get(index) - 1 : numbers.get(index))
+                .filter(n -> n % 2 == 1)
```
olekskov, line 40:
```suggestion
                .filter(number -> number % 2 == 1)
```

olekskov (approved):
Good job

PR #785, opened 2023-07-09
Solution

Stipaxa-hub:
Realized findMinEvenNumber method. This method return min integer value even numbers. If there is no needed data throw RuntimeException with message "Can't get min value from list: < Here is our input 'numbers' >".

Realized getOddNumsAverage method. This method return the average of all odd numbers from the list or throw NoSuchElementException, but before that subtract 1 from each element on an odd position. Realized selectMenByAge method. This method return list person which contains only men whose age is from 'fromAge' to 'toAge' inclusively. For example we can use this method for select men who can be recruited to army. Realized getWorkablePerson method. This method return list person which contains men whose age is from 'fromAge' and to 'maleToAge' and woman whose age is from 'fromAge' and to 'femaleToAge'. For example we can use this method for select people of working age. Realized getCatsName method. This method return list names of all cats whose owners are women from 'femaleAge' years old inclusively. Wrote implementation Predicate in CandidateValidator. Realized validateCandidates method. This method return all candidates whose are eligible to apply president position and return their names sorted alphabetically.

### DIFF
--- a/src/main/java/practice/CandidateValidator.java
+++ b/src/main/java/practice/CandidateValidator.java
@@ -1,5 +1,28 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.Arrays;
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MIN_OLD = 35;
+    private static final boolean ALLOWED_TO_VOTE = true;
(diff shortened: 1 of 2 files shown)


olekskov (approved):
Good job, approved

Please further avoid using 1-2 letters naming (like done now in findMinEvenNumber() ) even thing that looks simple today will be really hard to read in a month

PR #788, opened 2023-07-17
jv-stream-practice-solution

Sergey-Bogatkin:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,33 @@
...
+        }
+        if (!Objects.equals(candidate.getNationality(), ALLOWED_NATIONALITY)) {
+            return false;
+        }
+        String[] inUkraine = candidate.getPeriodsInUkr().split(DELIMITER);
+        int period = Integer.parseInt(inUkraine[TO]) - Integer.parseInt(inUkraine[FROM]);
+        if (period < MIN_ALLOWED_PERIOD) {
+            return false;
+        }
+        return true;
```
okuzan, line 31: combine all of these checks in one chain using &&

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,33 @@
...
+import java.util.Objects;
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    public static final int MIN_ALLOWED_AGE = 35;
+    public static final String ALLOWED_NATIONALITY = "Ukrainian";
+    public static final String DELIMITER = "-";
+    public static final int FROM = 0;
+    public static final int TO = 1;
```
okuzan, line 12: use more informative naming such as PERIOD_END_INDEX PERIOD_START_INDEX

okuzan (changes requested):

kozhukhovsky (approved):

PR #792, opened 2023-07-19
complate task

karvatskyi:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,24 @@
...
+                && candidate.getAge() >= MIN_YEAR_TO_RUN_FOR_PRESIDENT
+                && calculateYearsInUkraine(candidate) > 10
+                && candidate.isAllowedToVote();
+    }
+
+    private int calculateYearsInUkraine(Candidate candidate) {
+        String periodsInUkr = candidate.getPeriodsInUkr();
+        String[] years = periodsInUkr.split("-");
+        int startYear = Integer.parseInt(years[0]);
+        int endYear = Integer.parseInt(years[1]);
```
olekskov, line 21: Please extract magic values/indexes into constants with proper naming to increase code readability

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(n -> Arrays.stream(n.split(",")))
+                .map(Integer::parseInt)
+                .filter(n -> n % 2 == 0)
```
olekskov, line 26: Please avoid using 1-2 letter variable naming even in streams. After couple of weeks you won't remember what is "n", "s", "sb" and so on.
```suggestion
                .flatMap(stringValues -> Arrays.stream(stringValues.split(",")))
                .map(Integer::parseInt)
                .filter(number -> number % 2 == 0)
```
Please recheck whole solution for this issue

olekskov (changes requested):
good job overall, please improve codestyle

kozhukhovsky (approved):

PR #794, opened 2023-07-21
implemented new methods in StreamPractice

Den-moroz:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,30 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MIN_AGE = 35;
```
satori11111, line 7: do not shorten words

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,30 @@
...
+
+    private static final String ADMISSIBLE_NATIONALITY = "Ukrainian";
+    private static final int MIN_YEAR_IN_UKRAINE = 10;
```
satori11111, line 10: same

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,80 @@
...
-     * for example, `Arrays.asList( new Person(«Victor», 16, Sex.MAN),
-     * new Person(«Helen», 42, Sex.WOMAN))`,
-     * select from the List only men whose age is from `fromAge` to `toAge` inclusively.
-     * <p>
-     * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
-     */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> manFilter(person, fromAge, toAge))
```
satori11111, line 37: i think in common mistakes writes about creating Predicate not separate method, but you may be right

satori11111 (approved):
Good job, one little correction left

src/main/java/practice/StreamPractice.java
Den-moroz: I think in any case, it seems that creating a separate method  does not require creating a predicate in the method itself, but in principle, it might be better if create a predicate and not just a method

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,80 @@
...
-     * return the average of all odd numbers from the list or throw NoSuchElementException.
-     * But before that subtract 1 from each element on an odd position (having the odd index).
-     */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .mapToObj(i -> i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i))
+                .filter(num -> num % 2 == 1)
+                .mapToDouble(Integer::doubleValue)
+                .average()
```
satori11111, line 31: sorry, missed this in previous review. I think you do not need maptoObj and mapToDouble,you can do this with just map
              and average() already gives you OptionalDouble so you do not need to map it

satori11111 (changes requested):

satori11111 (approved):
awesome!!!

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,79 @@
...
-     * return min integer value. One more thing - we're interested in even numbers.
-     * If there is no needed data throw RuntimeException with message
-     * "Can't get min value from list: < Here is our input 'numbers' >"
-     */
+    private final CandidateValidator candidateValidator = new CandidateValidator();
+
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(s -> Arrays.stream(s.split(",")))
```
andrii-hoienko, line 18: what is 's'? Please make variable names more readable

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,79 @@
...
 
-    /**
-     * Given a List of Integer numbers,
-     * return the average of all odd numbers from the list or throw NoSuchElementException.
-     * But before that subtract 1 from each element on an odd position (having the odd index).
-     */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(i -> i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i))
```
andrii-hoienko, line 28: what is 'i'? Please make variable names more readable

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,79 @@
...
+                .filter(num -> num % 2 == 1)
```
andrii-hoienko, line 29: Code duplication (I mean it is better to make method wich will be filtering odd or even numbers and reuse it)

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,79 @@
...
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
andrii-hoienko, line 31: Please add some message here

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,79 @@
...
-     * Check which candidates are eligible to apply for president position and return their
-     * names sorted alphabetically.
-     * The requirements are: person should be older than 35 years, should be allowed to vote,
-     * have nationality - 'Ukrainian'
-     * and live in Ukraine for 10 years. For the last requirement use field periodsInUkr,
-     * which has following view: "2002-2015"
-     * We want to reuse our validation in future, so let's write our own impl of Predicate
-     * parametrized with Candidate in CandidateValidator.
-     */
     public List<String> validateCandidates(List<Candidate> candidates) {
```
andrii-hoienko, line 55: OPTIONAL: for consistensy it will be better to create separate validator for each case as you did for candidates

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,79 @@
...
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(candidateValidator)
+                .sorted(Comparator.comparing(Candidate::getName))
+                .map(Candidate::getName)
+                .collect(Collectors.toList());
+    }
+
+    private boolean manFilter(Person person, int fromAge, int toAge) {
```
andrii-hoienko, line 63: Bad method name, better to rename something like isValidMen, because mensFilter rather means that you consume a lot of mens and filter them, but is is not

andrii-hoienko (changes requested):

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,89 @@
...
+                && (person.getSex() == Person.Sex.MAN && person.getAge() <= maleToAge
+                || person.getSex() == Person.Sex.WOMAN && person.getAge() <= femaleToAge);
+    }
+
+    private boolean isValidWomenToHaveCat(Person person, int femaleAge) {
+        return person.getSex() == Person.Sex.WOMAN
+                && person.getAge() >= femaleAge;
+    }
+
+    private boolean isOddNumber(int number) {
```
andrii-hoienko, line 82: isOdd it is !IsEven, no need to define some logic here

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,89 @@
...
+                && person.getAge() <= toAge;
+    }
+
+    private boolean workable(Person person, int fromAge, int maleToAge, int femaleToAge) {
+        return person.getAge() >= fromAge
+                && (person.getSex() == Person.Sex.MAN && person.getAge() <= maleToAge
+                || person.getSex() == Person.Sex.WOMAN && person.getAge() <= femaleToAge);
+    }
+
+    private boolean isValidWomenToHaveCat(Person person, int femaleAge) {
```
andrii-hoienko, line 77: It this method we do not know anything about the cat, so I think it will be better to remove cat from method name

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,89 @@
...
+                .collect(Collectors.toList());
+    }
+
+    private boolean isValidMan(Person person, int fromAge, int toAge) {
+        return person.getSex() == Person.Sex.MAN
+                && person.getAge() > fromAge
+                && person.getAge() <= toAge;
+    }
+
+    private boolean workable(Person person, int fromAge, int maleToAge, int femaleToAge) {
```
andrii-hoienko, line 71: please rename to isWorkable for consistency

Vasul-Pankiv (approved):

PR #798, opened 2023-07-25
Implemented streampractice task

OlegTuryk:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,23 @@
...
+    private static final int MIN_YEARS_IN_UKRAINE = 10;
+    private static final String ELIGIBILITY_NATIONALITY = "Ukrainian";
+    private static final int START_YEAR_INDEX = 0;
+    private static final int END_YEAR_INDEX = 1;
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] periodInUkraine = candidate.getPeriodsInUkr().split(PERIOD_IN_UKRAINE_SEPARATOR);
+        int yearsInUkraine = Integer.parseInt(periodInUkraine[END_YEAR_INDEX])
+                - Integer.parseInt(periodInUkraine[START_YEAR_INDEX]);
```
Dimagaa, line 18: I would recommend you extract logic for receiving duration in a private method

PR #799, opened 2023-07-25
Homework

satori11111:


src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +72,12 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        Predicate<Person> predicate = person -> person.getSex() == Person.Sex.MAN
+                ? person.getAge() >= fromAge && person.getAge() <= maleToAge
+                : person.getAge() >= fromAge && person.getAge() <= femaleToAge;
```
Den-moroz, line 77: It's better, but it seems to me that it should be in a separate method, but it's already better than the previous one

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +86,14 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getSex() == Person.Sex.WOMAN
+                        && person.getAge() >= femaleAge)
```
Den-moroz, line 91: The same situation

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +86,14 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
...
+                .map(Person::getCats)
+                .flatMap(Collection::stream)
+                .map(Cat::getName)
```
Den-moroz, line 94: Here will be enough to use only flatMap and map

Den-moroz (changes requested):
Good solution! Let's improve) (if you can review me too)

Den-moroz (approved):
LGTM)

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(s -> Arrays.stream(s.split(",")))
```
andrii-hoienko, line 22: what is 's'? Please make variable names more readable

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,13 @@ public class StreamPractice {
...
+                .map(Integer::valueOf)
+                .filter(i -> i % 2 == 0)
```
andrii-hoienko, line 24: what is 'i'? Please make variable names more readable

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +33,12 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(i -> i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i))
+                .filter(integer -> integer % 2 != 0)
```
andrii-hoienko, line 38: code cuplication please reuse integer % 2 != 0 check

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +33,12 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(i -> i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i))
```
andrii-hoienko, line 37: what is 'i'? Please make variable names more readable

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +99,28 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
...
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(new CandidateValidator())
+                .map(Candidate::getName)
+                .sorted(String::compareTo)
+                .collect(Collectors.toList());
+    }
+
+    private boolean mensFilter(int fromAge, int toAge, Person person) {
```
andrii-hoienko, line 109: for consistensy it will be better to create separate validator for each case as you did for candidates

andrii-hoienko (changes requested):

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +99,28 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
...
+        return person.getAge() >= fromAge
+                && person.getAge() <= toAge
+                && person.getSex() == Person.Sex.MAN;
+    }
+
+    private static boolean workableFilter(int fromAge, int femaleToAge,
```
andrii-hoienko, line 115: why static?

src/main/java/practice/StreamPractice.java
andrii-hoienko, line 109: Bad method name, better to rename something like isValidMen, because mensFilter rather means that you consume a lot of mens and filter them, but is is not

satori11111: how we can create separate validator class if we have conditions that rely on parameters of method and method test in predicate can accept only one parameter?

andrii-hoienko: I mean not to use Predicate or create your own predicate. It is possible to create class PersonValidator or something and implement methods there, and then reuse in StreamPractice.java

andrii-hoienko (changes requested):
Comment about static methods is not resolved

andrii-hoienko (approved):
Respect

PR #802, opened 2023-07-26
Jv stream practice solution

keda-vlad:


Dimagaa (changes requested):
Good solution! just some suggestions for you

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,14 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(x -> x.split(","))
```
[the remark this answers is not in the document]
keda-vlad: Maybe, but its like string representation of int array, how it name shortly))

Dimagaa (approved):

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,14 @@ public class StreamPractice {
...
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(string -> string.split(","))
+                .flatMap(Arrays::stream)
+                .mapToInt(Integer::parseInt)
+                .filter(number -> number % 2 == 0)
+                .min()
+                .orElseThrow(() -> new RuntimeException("Can't get min value from list: "
+                        + numbers));
```
OlegTuryk, line 28: It's better to throw here NoSuchElementException

OlegTuryk (changes requested):

src/main/java/practice/StreamPractice.java
keda-vlad: Read the condition for this method: 

OlegTuryk (approved):

Vasul-Pankiv (approved):

PR #811, opened 2023-07-27
Wrote methods in StreamPractice

Someboty:


src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +36,14 @@ public int findMinEvenNumber(List<String> numbers) {
...
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(index -> (index % 2 == 1)
+                                ? numbers.get(index) - 1
+                                : numbers.get(index))
+                .filter(number -> number % 2 == 1)
+                .average()
+                .orElseThrow(() -> new NoSuchElementException(
```
AntonZhdanov, line 45: In this case, we have an overloaded method that throws NoSuchElementException by default
For example, you may have: .orElseThrow();
 or .getAsDouble();

AntonZhdanov (changes requested):

AntonZhdanov:
Nice work. Let's improve a bit.

AntonZhdanov (approved):

nacenik (approved):
LGTM! 👍

PR #826, opened 2023-07-31
solution for StreamPractice

vladis9:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,73 @@
...
+                .flatMap(s -> Arrays.stream(s.split(",")))
+                .map(Integer::parseInt)
+                .filter(this::isEven)
+                .collect(Collectors.toList());
+        if (integers.isEmpty()) {
+            throw new RuntimeException("Can't get min value from list: " + numbers);
+        }
+        return integers.stream()
+                .min(Integer::compare)
+                .get();
```
olekskov, line 25:
```suggestion
        return numbers.stream()
                .flatMap(s -> Arrays.stream(s.split(",")))
                .mapToInt(Integer::parseInt)
                .filter(this::isEven)
                .min()
                .orElseThrow(() -> new RuntimeException("Can't get min value from list: " + numbers))
```

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,73 @@
...
-    /**
-     * Given a List of `Person` instances (having `name`, `age`, `sex` and `cats` fields,
-     * and each `Cat` having a `name` and `age`),
-     * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
-     */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getSex() == Person.Sex.WOMAN)
+                .filter(person -> person.getAge() >= femaleAge)
```
olekskov, line 55: Please use one filter. If you feel that condition with && is complicated - introduce private boolean method like your isEven()

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,73 @@
...
-     * new Person(«Helen», 42, Sex.WOMAN))`,
-     * select from the List only men whose age is from `fromAge` to `toAge` inclusively.
-     * <p>
-     * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
-     */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getSex() == Person.Sex.MAN)
+                .filter(person -> person.getAge() >= fromAge && person.getAge() <= toAge)
```
olekskov, line 39: same here

olekskov (changes requested):

PR #830, opened 2023-08-06
Hw stream practice

kaptetos:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,25 @@
...
+    private boolean validatePeriodInUkraine(String periodsInUkr) {
+        String[] years = periodsInUkr.split("-");
+        if (years.length != 2) {
+            return false;
+        }
+
+        int startYear = Integer.parseInt(years[0]);
+        int endYear = Integer.parseInt(years[1]);
+        return endYear - startYear >= 10;
+    }
```
okuzan, line 24: 35,10, 0, 1, "-", Ukrainian should be constants

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +36,19 @@ public int findMinEvenNumber(List<String> numbers) {
...
+
+        List<Integer> oddNumbers = modifiedNumbers.stream()
+                .filter(num -> num % 2 != 0)
+                .collect(Collectors.toList());
+
+        if (oddNumbers.isEmpty()) {
+            throw new NoSuchElementException("No odd numbers in the list.");
+        }
+
+        return oddNumbers.stream().mapToInt(Integer::intValue).average().orElseThrow();
```
okuzan, line 51:
```suggestion
        return IntStream.range(...
```
group everything together, no need to use `if` and `mapToObj` and then `mapToInt`

okuzan (changes requested):

olekskov (approved):

PR #834, opened 2023-08-12
First

Olena-Pro:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,26 @@
...
+                >= ALLOWED_LIVE_IN;
+        return candidate.getAge() >= ALLOWED_AGE
+                && candidate.getNationality().equals(PERSON_NATIONALITY)
+                && period && candidate.isAllowedToVote();
+    }
 }
+
+//write your code here
+
+
```
olekskov, line 26: unnecessary lines:
```suggestion
```

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,26 @@
...
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final String PERSON_NATIONALITY = "Ukrainian";
+    private static final int INDEX_0 = 0;
+    private static final int INDEX_1 = 1;
```
olekskov, line 9: There is no any additional information in such naming. Why 0 and why 1 new person will never guess easily
```suggestion
    private static final int INDEX_DATE_FROM = 0;
    private static final int INDEX_DATE_TO = 1;
```

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +47,11 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getAge() >= fromAge
+                        && p.getAge() <= toAge
+                        && p.getSex().equals(Person.Sex.MAN))
```
olekskov, line 53: Please replace all your s, n, p etc with proper full name. After 2 weeks you will not remember what that abbreviation means
```suggestion
                .filter(person -> person.getAge() >= fromAge
                        && person.getAge() <= toAge
                        && person.getSex().equals(Person.Sex.MAN))
```

olekskov (changes requested):

olekskov (approved):
Good job

PR #839, opened 2023-08-16
hw-stream-github-practice

tanneken:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(num -> Arrays.stream(num.split(",")))
```
Ivan95kos, line 23: All magic numbers, strings, and symbols in your code should be constants.
don't use short names, fix them everywhere
```suggestion
                .flatMap(line -> Arrays.stream(line.split(SEPARATOR)))
```

Ivan95kos (changes requested):
Good job overall!

olekskov (approved):
Good job

PR #843, opened 2023-08-23
stream solution

margosha99:


okuzan (changes requested):
Please make sure the build passes before re-requesting a review. Also, all literals- indexes, strings, should be constants. Don't use if in stream chains, you can use a ternary operator. Don't use short variables' names with one letter, better name them `number` etc.

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,21 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+
```
kozhukhovsky, line 7:
```suggestion
```

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +34,12 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        Double sum = IntStream
+                .range(0, numbers.size())
+                .map(index -> index % 2 != 0 ? numbers.get(index) - 1 : numbers.get(index))
+                .filter(number -> number % 2 != 0)
+                .average().getAsDouble();
```
kozhukhovsky, line 41:
```suggestion
                .average()
                .getAsDouble();
```

kozhukhovsky (approved):

PR #844, opened 2023-08-23
Initial commit

gordgelin:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,39 @@
...
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final String NATIONALITY = "Ukrainian";
+    private static final String YEARS_DELIMITER = "-";
+    private static final int LOWER_AGE_LIMIT = 35;
+    private static final int MIN_RESIDENCY_YEARS = 10;
+    private static final int FROM_YEAR = 0;
+    private static final int TO_YEAR = 1;
```
okuzan, line 12: Well done with extracting to constants, but this isn't a year, it's more of an index of the year :)
```suggestion
    private static final int FROM_YEAR_INDEX = 0;
    private static final int TO_YEAR_INDEX  = 1;
```

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,39 @@
...
+
+        if (!NATIONALITY.equals(candidate.getNationality())) {
+            return false;
+        }
+
+        if (candidate.getAge() < LOWER_AGE_LIMIT) {
+            return false;
+        }
+
+        return true;
```
okuzan, line 36: in this case, we can achieve the same result more elegantly, without ifs - just chain all of these conditions with && - this is fail fast operator, when we will get the FALSE, chain stops and returns false - the same logic as you have here

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,12 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMapToInt(s -> Stream.of(s.split(",")).mapToInt(Integer::valueOf))
+                .filter(i -> i % 2 == 0)
```
okuzan, line 23: better to not use short one-letter naming even in streams, we decrease readability and don't really achieve anything other than a few seconds to type out longer name :) remember that developers spend more time reading code of others, than writing their own

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +32,16 @@ public int findMinEvenNumber(List<String> numbers) {
...
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .mapToDouble(i -> {
+                    if (i % 2 != 0) {
+                        return numbers.get(i) - 1;
+                    }
+                    return numbers.get(i);
+                })
```
okuzan, line 41: ternary operator will make it more concise

okuzan (changes requested):

src/main/java/practice/StreamPractice.java
gordgelin: i had to short names to one-letter values due to stylecheck line length limitation of 100
there is not always a way to put some fine looking line break...

okuzan (changes requested):
rename p to person etc, break the lines if checkstyle requires to do so

okuzan (approved):
Good job

PR #854, opened 2023-09-05
implemented all tasks

teract10s:


src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +74,15 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        Predicate<Person> peopleFromAgeToAge = person -> {
+            if (person.getSex() == Person.Sex.MAN) {
+                return person.getAge() >= fromAge && person.getAge() <= maleToAge;
+            }
+            return person.getAge() >= fromAge && person.getAge() <= femaleToAge;
+        };
```
liliia-ponomarenko, line 82: better use || and make it inline lambda

liliia-ponomarenko (approved):
Great 😊

src/main/java/practice/StreamPractice.java
teract10s: I thought about it. However, it looked clearer to me, I don't know why

PR #856, opened 2023-09-07
StreamPractice is implemented

Alexsandr-Yablunovskyi:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,18 @@
...
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int REQUIRED_AGE = 35;
+    private static final int REQUIRED_PERIOD_IN_UKR = 10;
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] years = candidate.getPeriodsInUkr().split("-");
+        int periodInUkr = Integer.parseInt(years[1]) - Integer.parseInt(years[0]);
+        return candidate.getAge() >= REQUIRED_AGE && candidate.isAllowedToVote()
+                && candidate.getNationality().equals("Ukrainian")
```
don-bigdad, line 15: I created separate methods for each check, but okay

src/main/java/practice/StreamPractice.java
```diff
@@ -13,8 +17,19 @@ public class StreamPractice {
...
-        return 0;
+    public static int findMinEvenNumber(List<String> numbers) {
+        List<Integer> list = numbers.stream()
+                .map(i -> i.split(","))
+                .flatMap(i -> Arrays.stream(i).distinct())
+                .map(Integer::parseInt)
+                .filter(i -> i % 2 == 0)
+                .sorted()
+                .limit(1)
+                .toList();
```
don-bigdad, line 28: mapToInt have method min() which gave you minimal value of stream

src/main/java/practice/StreamPractice.java
```diff
@@ -13,8 +17,19 @@ public class StreamPractice {
...
+        if (list.size() == 0) {
+            throw new RuntimeException("Can't get min value from list: " + numbers);
+        }
```
don-bigdad, line 31: I think mentors want to see the use of orElseThrow

don-bigdad (approved):

src/main/java/practice/CandidateValidator.java
liliia-ponomarenko, line 15: separator, nationality, and indexes also should  make constant

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +34,16 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(i -> {
+                    if (i % 2 == 1) {
+                        return numbers.get(i) - 1;
+                    }
+                    return numbers.get(i);
```
liliia-ponomarenko, line 42: Let’s  use a ternary operator

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +34,16 @@ public int findMinEvenNumber(List<String> numbers) {
...
+                })
+                .filter(i -> i % 2 == 1)
+                .average()
+                .orElseThrow();
```
liliia-ponomarenko, line 46: you can use getAsDouble()

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +73,17 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
...
+        Predicate<Person> isWorkable = i -> {
+            if (i.getSex() == Person.Sex.MAN) {
+                return i.getAge() <= maleToAge;
+            } else {
+                return i.getAge() <= femaleToAge;
+            }
+        };
+        return peopleList.stream()
+                .filter(i -> i.getAge() >= fromAge)
+                .filter(isWorkable)
```
liliia-ponomarenko, line 85: you can combine these filters

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +92,13 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(i -> i.getSex() == Person.Sex.WOMAN && i.getAge() >= femaleAge)
+                .map(Person::getCats)
+                .flatMap(Collection::stream)
+                .map(Cat::getName)
+                .distinct()
```
liliia-ponomarenko, line 100: do you really  need .distinct() here?

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +114,10 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(i -> new CandidateValidator().test(i))
```
liliia-ponomarenko, line 118: you can use method reference  instead

liliia-ponomarenko (changes requested):
Good job! Let’s  improve your solution ;)

src/main/java/practice/StreamPractice.java
Alexsandr-Yablunovskyi: I did, but IDEA threw a message, that orElseThrow() is prefered method for this situation, so I changed it.

liliia-ponomarenko (approved):
Great! 😊

PR #858, opened 2023-09-08
Impemented CandidateValidator.test, all StreamPractice methods were a…

torbinofficial:
…lso implemented

### DIFF
--- a/src/main/java/practice/CandidateValidator.java
+++ b/src/main/java/practice/CandidateValidator.java
@@ -1,5 +1,19 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= 35
+                && candidate.getNationality().equals("Ukrainian")
+                && validatePeriod(candidate)
+                && candidate.isAllowedToVote();
+    }
+
+    private boolean validatePeriod(Candidate candidate) {
+        String[] years = candidate.getPeriodsInUkr().split("-");
+        return Integer.parseInt(years[1]) - Integer.parseInt(years[0]) > 10;
+    }
 }
--- a/src/main/java/practice/StreamPractice.java
+++ b/src/main/java/practice/StreamPractice.java
@@ -1,8 +1,11 @@
 package practice;
 
-import java.util.Collections;
+import java.util.Arrays;
 import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.stream.IntStream;
 import model.Candidate;
+import model.Cat;
 import model.Person;
 
 public class StreamPractice {
@@ -14,7 +17,15 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMapToInt(x -> Arrays.stream(x.split(","))
+                        .mapToInt(Integer::valueOf))
+                .filter(x -> x % 2 == 0)
+                .min()
+                .orElseThrow(() -> new RuntimeException(
+                        "Can't get min value from list: "
+                                + numbers
+                ));
     }
 
     /**
@@ -23,7 +34,11 @@ public class StreamPractice {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(number -> numbers.get(number) - number % 2)
+                .filter(num -> num % 2 != 0)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
     }
 
     /**
@@ -35,7 +50,11 @@ public class StreamPractice {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(x -> x.getSex().equals(Person.Sex.MAN)
+                        && fromAge <= x.getAge()
+                        && x.getAge() <= toAge)
+                .toList();
     }
 
     /**
@@ -50,7 +69,11 @@ public class StreamPractice {
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(x -> fromAge <= x.getAge()
+                        && (x.getSex().equals(Person.Sex.MAN)
+                        ? (x.getAge() <= maleToAge) : (x.getAge() <= femaleToAge)))
+                .toList();
     }
 
     /**
@@ -59,22 +82,31 @@ public class StreamPractice {
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(x -> x.getSex().equals(Person.Sex.WOMAN)
+                        && femaleAge < x.getAge())
+                .flatMap(x -> x.getCats().stream().map(Cat::getName))
+                .toList();
     }
 
     /**
-     * Your help with a election is needed. Given list of candidates, where each element
+     * Your help with an election is needed. Given list of candidates, where each element
      * has Candidate.class type.
      * Check which candidates are eligible to apply for president position and return their
      * names sorted alphabetically.
      * The requirements are: person should be older than 35 years, should be allowed to vote,
      * have nationality - 'Ukrainian'
      * and live in Ukraine for 10 years. For the last requirement use field periodsInUkr,
-     * which has following view: "2002-2015"
-     * We want to reuse our validation in future, so let's write our own impl of Predicate
+     * which has the following view: "2002-2015"
+     * We want to reuse our validation in the future, so let's write our own impl of Predicate
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        CandidateValidator candidateValidator = new CandidateValidator();
+        return candidates.stream()
+                .filter(candidateValidator)
+                .map(Candidate::getName)
+                .sorted()
+                .toList();
     }
 }


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,19 @@
...
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= 35
```
VolodymyrSkvira, line 9: You should use constants for your checks

```suggestion
public class CandidateValidator implements Predicate<Candidate> {
    private static final int  PRESIDENT_MIN_AG = 35;
    
    @Override
    public boolean test(Candidate candidate) {
        return candidate.getAge() >= PRESIDENT_MIN_AGE;
```

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,19 @@
...
+                && candidate.getNationality().equals("Ukrainian")
+                && validatePeriod(candidate)
+                && candidate.isAllowedToVote();
+    }
+
+    private boolean validatePeriod(Candidate candidate) {
+        String[] years = candidate.getPeriodsInUkr().split("-");
+        return Integer.parseInt(years[1]) - Integer.parseInt(years[0]) > 10;
+    }
```
VolodymyrSkvira, line 18: This method`s body can be moved to the test method above that you override

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,15 @@ public class StreamPractice {
...
-        return 0;
+        return numbers.stream()
+                .flatMapToInt(x -> Arrays.stream(x.split(","))
+                        .mapToInt(Integer::valueOf))
+                .filter(x -> x % 2 == 0)
+                .min()
+                .orElseThrow(() -> new RuntimeException(
+                        "Can't get min value from list: "
+                                + numbers
+                ));
```
VolodymyrSkvira, line 29: Minor formating mishap

```suggestion
        return numbers.stream()
                .flatMapToInt(x -> Arrays.stream(x.split(","))
                .mapToInt(Integer::valueOf))
                .filter(x -> x % 2 == 0)
                .min()
                .orElseThrow(() -> new RuntimeException(
                        "Can't get min value from list: "
                                + numbers
                ));
```

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +51,14 @@ public Double getOddNumsAverage(List<Integer> numbers) {
...
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        Predicate<Person> personPredicate = (person) ->
+                person.getSex().equals(Person.Sex.MAN)
+                && fromAge <= person.getAge()
+                && person.getAge() <= toAge;
+
+        return peopleList.stream()
+                .filter(personPredicate)
+                .toList();
```
VolodymyrSkvira, line 61: Predicate can be avoided here but it is up to you. IMO it is much more needed in the next task

VolodymyrSkvira (changes requested):
Great solution! Just some small issues

src/main/java/practice/StreamPractice.java
torbinofficial: We can let a mentor to decide

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,18 @@
...
+    private static final int MINIMUM_AGE = 35;
+    private static final String REQUIRED_NATIONALITY = "Ukrainian";
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] years = candidate.getPeriodsInUkr().split("-");
+        return candidate.getAge() >= MINIMUM_AGE
+                && candidate.getNationality().equals(REQUIRED_NATIONALITY)
+                && candidate.isAllowedToVote()
+                && Integer.parseInt(years[1]) - Integer.parseInt(years[0]) > 10;
```
liliia-ponomarenko, line 16: period of living in Ukraine, separator, and indexes also better make constant ;)

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +35,11 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(number -> numbers.get(number) - number % 2)
+                .filter(num -> num % 2 != 0)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
liliia-ponomarenko, line 42: you can use getAsDouble here

src/main/java/practice/StreamPractice.java
liliia-ponomarenko: I like solutions with local variables for a predicate. It looks more readable. Would be great if the name of the variable informed which  predicate applying  (isPersonApplicableToArmy)

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +73,14 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        Predicate<Person> personPredicate = person ->
```
liliia-ponomarenko, line 76: the same about name

src/main/java/practice/StreamPractice.java
```diff
@@ -59,22 +89,34 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
...
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        Predicate<Person> personPredicate = (person) ->
+                person.getSex().equals(Person.Sex.WOMAN)
+                && femaleAge < person.getAge();
+
+        return peopleList.stream()
+                .filter(personPredicate)
+                .flatMap(x -> x.getCats().stream().map(Cat::getName))
```
liliia-ponomarenko, line 98:
```suggestion
                .flatMap(x -> x.getCats().stream()
                      .map(Cat::getName))
```

liliia-ponomarenko (changes requested):
Good job! A few recommendations that help to improve your solution ;)

liliia-ponomarenko (changes requested):
;)

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,20 @@
...
+    private static final int MINIMUM_YEARS = 10;
+    private static final String SEPARATOR = "-";
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] years = candidate.getPeriodsInUkr().split(SEPARATOR);
+        return candidate.getAge() >= MINIMUM_AGE
+                && candidate.getNationality().equals(REQUIRED_NATIONALITY)
+                && candidate.isAllowedToVote()
+                && Integer.parseInt(years[1]) - Integer.parseInt(years[0]) > MINIMUM_YEARS;
```
NazarSmith228, line 18: would make sense to create separate variables for start of the period and end

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +34,11 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(number -> numbers.get(number) - number % 2)
+                .filter(num -> num % 2 != 0)
```
NazarSmith228, line 39: minor

in one lambda you use `number` in other - `num`

a good practice is to have parameter names as short as possible - you could use `n` in both cases

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +34,11 @@ public int findMinEvenNumber(List<String> numbers) {
...
+                .average()
+                .getAsDouble();
```
NazarSmith228, line 41: you can use this method, although a better option here would be `orElseThrow`

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +50,14 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        Predicate<Person> isPersonApplicableToArmy = (person) ->
+                person.getSex().equals(Person.Sex.MAN)
```
NazarSmith228, line 54: no need to use `equals` when comparing enums - you can use `==` instead

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +72,14 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        Predicate<Person> isPersonCanWork = person ->
+                fromAge <= person.getAge()
+                        && (person.getSex().equals(Person.Sex.MAN)
+                        ? (person.getAge() <= maleToAge) : (person.getAge() <= femaleToAge));
```
NazarSmith228, line 78:
```suggestion
        Predicate<Person> personCanWork = person ->
                fromAge <= person.getAge()
                        && (person.getSex().equals(Person.Sex.MAN)
                        ? person.getAge() <= maleToAge : person.getAge() <= femaleToAge);
```

src/main/java/practice/StreamPractice.java
```diff
@@ -59,22 +88,35 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
...
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        Predicate<Person> isItWomanWithCats = (person) ->
+                person.getSex().equals(Person.Sex.WOMAN)
+                && femaleAge < person.getAge();
+
+        return peopleList.stream()
+                .filter(isItWomanWithCats)
+                .flatMap(x -> x.getCats().stream()
+                        .map(Cat::getName))
```
NazarSmith228, line 98:
```suggestion
                .flatMap(x -> x.getCats().stream())
                .map(Cat::getName)
```

NazarSmith228 (changes requested):

liliia-ponomarenko (approved):
LGTM 🔥

PR #861, opened 2023-09-09
HW stream github practice

OlehReex:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,15 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(string -> string.split(","))
+                .flatMap(Arrays::stream)
+                .mapToInt(Integer::parseInt)
+                .boxed()
```
sonyasapsan, line 25: you don't need to call boxed(). remove it please

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +69,14 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> {
+                    if (person.getSex().equals(Person.Sex.MAN)) {
+                        return person.getAge() >= fromAge && person.getAge() <= maleToAge;
```
sonyasapsan, line 75: I'm not sure,  but what about ternary operator? then you don't need to create the lambda expression with body

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +107,17 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
...
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(candidate -> candidate.getAge() >= 35)
+                .filter(candidate -> candidate.getNationality().equals("Ukrainian"))
+                .filter(Candidate::isAllowedToVote)
+                .filter(candidate -> {
+                    String[] periodString = candidate.getPeriodsInUkr().split("-");
+                    return Integer.parseInt(periodString[1])
+                            - Integer.parseInt(periodString[0]) >= 10;
+                })
```
sonyasapsan, line 118: you should check a candidate in CandidateValidator.java, not here

src/main/java/practice/StreamPractice.java
OlehReex: Yeah, good idea

sonyasapsan (approved):

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +105,10 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(candidateValidator)
+                .map(Candidate::getName)
+                .sorted()
+                .toList();
     }
```
evgenkolner, line 113: as for me solution is perfect

evgenkolner (approved):

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,23 @@
...
+    public boolean test(Candidate candidate) {
+        return candidate.getNationality().equals(NATIONALITY)
+                && candidate.isAllowedToVote()
+                && candidate.getAge() >= VALID_AGE
+                && getYearsInUkr(candidate) >= VALID_PERIOD;
+    }
+
+    private int getYearsInUkr(Candidate candidate) {
+        String[] periodString = candidate.getPeriodsInUkr().split("-");
+        return Integer.parseInt(periodString[1]) - Integer.parseInt(periodString[0]);
```
liliia-ponomarenko, line 21: Separators and indexes also should make constant

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,14 @@ public class StreamPractice {
...
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(string -> string.split(","))
+                .flatMap(Arrays::stream)
+                .mapToInt(Integer::parseInt)
+                .filter(integer -> integer % 2 == 0)
+                .sorted()
+                .findFirst()
```
liliia-ponomarenko, line 29:
```suggestion
                .min()
```
You can use min() instead

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,14 @@ public class StreamPractice {
...
+                .orElseThrow(() -> new RuntimeException("Can't get min value from list"));
```
liliia-ponomarenko, line 30: ["Can't get min value from list: " + numbers](https://mate-academy.github.io/jv-program-common-mistakes/java-core/java-eight-part-two/stream-api#in-method-findminevennumber-exception-message-should-look-like-this)

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +36,11 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(i -> i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i))
+                .filter(number -> number % 2 != 0)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
liliia-ponomarenko, line 43: [If you need to throw exception when no result found after stream execution sometimes get()/getAsDouble() may help, try to google what does it do.](https://mate-academy.github.io/jv-program-common-mistakes/java-core/java-eight-part-two/stream-api#try-to-use-a-minimum-amount-of-stream-operations)

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +52,10 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getSex().equals(Person.Sex.MAN))
+                .filter(person -> person.getAge() >= fromAge && person.getAge() <= toAge)
```
liliia-ponomarenko, line 57: you can combine these filtration. Also, big lambda expression  better move from stream to local variable (apply it for other methods too)

liliia-ponomarenko (changes requested):
Good job! A few recommendations that help to improve your solution ;)

liliia-ponomarenko (approved):
LGTM 🔥

PR #863, opened 2023-09-10
First version, check please

Alex-zzzz:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,20 @@
...
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MIN_AGE = 35;
+    private static final int MIN_PERIOD_LIVE_IN_UKRAINE = 10;
+
+    @Override
+    public boolean test(Candidate candidate) {
+
+        boolean isPeriodOk = Integer.valueOf(candidate.getPeriodsInUkr().substring(5))
+                - Integer.valueOf(candidate.getPeriodsInUkr().substring(0,4))
+                >= MIN_PERIOD_LIVE_IN_UKRAINE;
```
okuzan, line 15: better split by separator

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,20 @@
...
+
+        return candidate.getAge() >= MIN_AGE && candidate.isAllowedToVote()
+                && candidate.getNationality() == "Ukrainian" && isPeriodOk;
```
okuzan, line 18: ukrainian should be constant

okuzan (changes requested):
Rename short one-letter variables in stream chains to something more readable, please :)

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +16,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+
+        return numbers.stream().map(e -> e.split(","))
```
Ivan95kos, line 20: For the future, don't use short names, fix them for yourself
All magic numbers, strings, and symbols in your code should be constants.
```suggestion
        return numbers.stream().map(line -> line.split(SEPARATOR))
```
It looks more readable

Ivan95kos (approved):

PR #864, opened 2023-09-10
I have implemented all the methods

Cayman4010:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,20 @@
...
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int FIRST_YEAR = 0;
+    private static final int LAST_YEAR = 1;
+    private static final int MIN_TERM = 10;
+    private static final int MIN_AGE = 35;
+
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= MIN_AGE
+                && candidate.getNationality().equals("Ukrainian")
```
Vt0rtikV, line 15: Maybe "Ukrainian" can be constant too

src/main/java/practice/StreamPractice.java
```diff
@@ -1,20 +1,34 @@
...
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(string -> string.split(NUMBER_SEPARATOR))
+                .flatMap(Arrays::stream)
+                .map(Integer::parseInt)
+                .filter(number -> number % 2 == 0)
+                .min(Comparator.naturalOrder()).orElseThrow(()
+                        -> new RuntimeException("Can't get min value from list: "
+                        + "< Here is our input 'numbers' >"));
```
Vt0rtikV, line 31: May be like this it would be better)
```suggestion
                        -> new RuntimeException("Can't get min value from list: "
                        + numbers));
```

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +52,11 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
```
Vt0rtikV, line 52: I think you don't need this comments

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +52,11 @@ public Double getOddNumsAverage(List<Integer> numbers) {
...
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getAge() >= fromAge
+                        && person.getAge() <= toAge
+                        && person.getSex() == Person.Sex.MAN)
```
Vt0rtikV, line 58: Maybe it will be better in Predicate before return

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +71,11 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getSex() == Person.Sex.MAN
+                        ? person.getAge() >= fromAge && person.getAge() <= maleToAge
+                        : person.getAge() >= fromAge && person.getAge() <= femaleToAge)
```
Vt0rtikV, line 77: The same in previous method

Vt0rtikV (changes requested):

src/main/java/practice/StreamPractice.java
Cayman4010: I think, it doesn`t matter in this situation

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,21 @@
...
+    private static final int MIN_AGE = 35;
+    private static final String VALID_NATIONALITY = "Ukrainian";
+
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= MIN_AGE
+                && candidate.getNationality().equals(VALID_NATIONALITY)
+                && candidate.isAllowedToVote()
+                && Integer.parseInt(candidate.getPeriodsInUkr().split("-")[LAST_YEAR])
+                - Integer.parseInt(candidate.getPeriodsInUkr().split("-")[FIRST_YEAR]) >= MIN_TERM;
```
liliia-ponomarenko, line 19: candidate.getPeriodsInUkr().split("-") - let's make it a local variable.
"-" - let's make it constant

src/main/java/practice/StreamPractice.java
```diff
@@ -1,20 +1,33 @@
...
      */
+
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(string -> string.split(NUMBER_SEPARATOR))
+                .flatMap(Arrays::stream)
+                .map(Integer::parseInt)
+                .filter(number -> number % 2 == 0)
+                .min(Comparator.naturalOrder()).orElseThrow(()
```
liliia-ponomarenko, line 28:
```suggestion
                .min()
                .orElseThrow(()
```
It’s  default  behavior min()

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +36,10 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(number -> number % 2 == 1 ? numbers.get(number) - 1 : numbers.get(number))
+                .filter(number -> number % 2 == 1)
+                .average().orElseThrow(NoSuchElementException::new);
```
liliia-ponomarenko, line 42:
```suggestion
                .average()
                .getAsDouble;
```
you can use getAsDouble

src/main/java/practice/StreamPractice.java
liliia-ponomarenko: agree, big lambda expression better move from stream to local variable (apply it for other methods too)

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +104,10 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(new CandidateValidator())
+                .map(Candidate::getName)
+                .sorted(Comparator.naturalOrder())
```
liliia-ponomarenko, line 110: the same

liliia-ponomarenko (changes requested):
Good job! Let’s  improve  your solution;)

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +101,28 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
...
```
liliia-ponomarenko, line 107:
```suggestion
                .sorted()
```

liliia-ponomarenko (approved):
good job! 👍

PR #869, opened 2023-09-11
implemented all the methods from streamPractice. Implemented Candidat…

vsych321:
…eValidator

### DIFF
--- a/src/main/java/practice/StreamPractice.java
+++ b/src/main/java/practice/StreamPractice.java
@@ -1,11 +1,17 @@
 package practice;
 
-import java.util.Collections;
+import java.util.Arrays;
 import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import model.Candidate;
+import model.Cat;
 import model.Person;
 
 public class StreamPractice {
+    private static final int SUBTRACT1 = 1;
     /**
      * Given list of strings where each element contains 1+ numbers:
      * input = {"5,30,100", "0,22,7", ...}
@@ -13,8 +19,16 @@ public class StreamPractice {
      * If there is no needed data throw RuntimeException with message
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
+
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(string -> string.split(","))
+                .flatMap(Arrays::stream)
+                .mapToInt(Integer::parseInt)
+                .filter(number -> number % 2 == 0)
+                .min()
+                .orElseThrow(() -> new RuntimeException("Can't get min value from list: "
+                        + numbers));
     }
 
     /**
@@ -23,7 +37,13 @@ public class StreamPractice {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .mapToDouble(oddNumber -> oddNumber % 2 != 0 ? numbers.get(oddNumber) - SUBTRACT1
+                        : numbers.get(oddNumber))
+                .filter(oddIndex -> oddIndex % 2 != 0)
+                .average()
+                .orElseThrow(() -> new NoSuchElementException("Can't get average of odds from "
+                        + numbers));
     }
 
     /**
@@ -35,7 +55,10 @@ public class StreamPractice {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getSex() == Person.Sex.MAN
+                && person.getAge() >= fromAge && person.getAge() <= toAge)
+                .collect(Collectors.toList());
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +55,10 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getSex() == Person.Sex.MAN
+                && person.getAge() >= fromAge && person.getAge() <= toAge)
```
Parashchuk, line 60: I'm not sure, but it's may be better to create a new Predicate if there's more than two line in condition

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +88,13 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getSex() == Person.Sex.WOMAN
+                && person.getAge() >= femaleAge && !person.getCats().isEmpty())
```
Parashchuk, line 93: The same here about two lines

Parashchuk (approved):

src/main/java/practice/StreamPractice.java
```diff
@@ -1,20 +1,34 @@
...
+import java.util.NoSuchElementException;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import model.Candidate;
+import model.Cat;
 import model.Person;
 
 public class StreamPractice {
+    private static final int SUBTRACT1 = 1;
```
NazarSmith228, line 14: weird constant

it makes no sense, you can omit that

otherwise you would need to create constants for each tiny thing

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +37,13 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .mapToDouble(oddNumber -> oddNumber % 2 != 0 ? numbers.get(oddNumber) - SUBTRACT1
```
NazarSmith228, line 41: you can rename `oddNumber` to `n` and make the lambda shorter

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +37,13 @@ public int findMinEvenNumber(List<String> numbers) {
...
+                        : numbers.get(oddNumber))
+                .filter(oddIndex -> oddIndex % 2 != 0)
```
NazarSmith228, line 43: this isn't actually an index

common approach is to name lambda parameters as short as possible (`n` in this case)

src/main/java/practice/StreamPractice.java
NazarSmith228: we can actually use 2 separate `filter` statements

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +55,10 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getSex() == Person.Sex.MAN
+                && person.getAge() >= fromAge && person.getAge() <= toAge)
+                .collect(Collectors.toList());
```
NazarSmith228, line 61: `toList()`

**apply to all places**

NazarSmith228 (changes requested):

src/main/java/practice/StreamPractice.java
Parashchuk: Won't it cost more performance to make two loops over whole data structure ?

NazarSmith228: not in this case definitely 

performance in terms of Streams is not dependent on some underlying structures (no structures are used)

instead it depends on Stream characteristic (ORDERED, SORTED, DISTINCT etc)

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +86,13 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
...
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getSex() == Person.Sex.WOMAN
+                && person.getAge() >= femaleAge && !person.getCats().isEmpty())
+                .flatMap(catsArray -> catsArray.getCats().stream())
+                .map(Cat::getName)
+                .toList();
+
```
liliia-ponomarenko, line 95:
```suggestion
```

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +108,11 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(new CandidateValidator())
+                .map(Candidate::getName)
+                .sorted()
+                .toList();
+
```
liliia-ponomarenko, line 116:
```suggestion
```

PR #871, opened 2023-09-11
complete all tastks from stream Practice

anigilator479:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,23 @@
...
+    private static final String CANDIDATE_REQUIRED_NATIONALITY = "Ukrainian";
+    private static final int CANDIDATE_REQUIRED_YEARS_IN_UKRAINE = 10;
+    private static final boolean ALLOWED_TO_VOTE = true;
+    private static final String SPLITERATOR = "-";
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] periodInUkraine = candidate.getPeriodsInUkr().split(SPLITERATOR);
+        int candidateYearsInUkraine = Integer.parseInt(periodInUkraine[1])
+                - Integer.parseInt(periodInUkraine[0]);
```
TetianaDenchyk, line 17: You may put into sepatate private method

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,23 @@
...
+        return candidate.getAge() >= CANDIDATE_REQUIRED_AGE
+                && candidate.getNationality().equals(CANDIDATE_REQUIRED_NATIONALITY)
+                && candidate.isAllowedToVote() == ALLOWED_TO_VOTE
```
TetianaDenchyk, line 20: I think the constant `ALLOWED_TO_VOTE` is redundant. When you call `candidate.isAllowedToVote()` it already gives you necessary result

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,14 @@ public class StreamPractice {
...
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(num -> Arrays.stream(num.split(",")))
+                .map(Integer::parseInt)
+                .filter(num -> num % 2 == 0)
+                .sorted()
+                .findFirst()
+                .orElseThrow(() -> new RuntimeException("Can't get min value from list"
+                        + ": method_input_list"));
```
TetianaDenchyk, line 28: I am not quite sure what `method_input_list` refers to here but I guess you should mention here the input list

TetianaDenchyk (approved):
Great job :) I left several comments for you to fix if you agree with them

src/main/java/practice/CandidateValidator.java
liliia-ponomarenko: agree

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,14 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(num -> Arrays.stream(num.split(",")))
+                .map(Integer::parseInt)
+                .filter(num -> num % 2 == 0)
+                .sorted()
+                .findFirst()
```
liliia-ponomarenko, line 26: you can use min()
```suggestion
                .min()
```

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,14 @@ public class StreamPractice {
...
+                .orElseThrow(() -> new RuntimeException("Can't get min value from list"
+                        + ":" + numbers));
```
liliia-ponomarenko, line 28:
```suggestion
                .orElseThrow(() -> new RuntimeException("Can't get min value from list: "
                        + numbers));
```

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +52,10 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(human -> human.getAge() >= fromAge
+                        && human.getAge() <= toAge && human.getSex().equals(Person.Sex.MAN))
```
liliia-ponomarenko, line 57: with enums you can use == 
also, large lambda expression  better move to separate  variable

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +52,10 @@ public Double getOddNumsAverage(List<Integer> numbers) {
...
+                .collect(Collectors.toList());
```
liliia-ponomarenko, line 58:
```suggestion
                .toList();
```
Let’s  apply it to all methods

liliia-ponomarenko (changes requested):
Good job! Let’s improve your solution. But be attentive and check if other methods can have the same issue;)

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +85,14 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        Predicate<Person> personPredicate = human -> human.getSex() == Person.Sex.WOMAN
+                && human.getAge() >= femaleAge;
+        return peopleList.stream()
+                .filter(personPredicate)
+                .map(Person::getCats)
+                .flatMap(List::stream)
```
Andriy-dz, line 93: you can combine these methods

Andriy-dz (approved):

PR #872, opened 2023-09-11
solved stream practice

TetianaDenchyk:


sonyasapsan (approved):
Looks good to me! Well done)

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,62 @@
...
-     * for example, `Arrays.asList( new Person(«Victor», 16, Sex.MAN),
-     * new Person(«Helen», 42, Sex.WOMAN))`,
-     * select from the List only men whose age is from `fromAge` to `toAge` inclusively.
-     * <p>
-     * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
-     */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getSex().equals(Person.Sex.MAN)
```
liliia-ponomarenko, line 33: You can use == for comparing enums

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,62 @@
...
-     * have nationality - 'Ukrainian'
-     * and live in Ukraine for 10 years. For the last requirement use field periodsInUkr,
-     * which has following view: "2002-2015"
-     * We want to reuse our validation in future, so let's write our own impl of Predicate
-     * parametrized with Candidate in CandidateValidator.
-     */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(c -> new CandidateValidator().test(c))
```
liliia-ponomarenko, line 57:
```suggestion
                .filter(new CandidateValidator())
```
you can simplify  it

liliia-ponomarenko (changes requested):
Good job! Let’s improve your solution. But be attentive and check if other methods can have the same issue;)

Andriy-dz (approved):

PR #876, opened 2023-09-12
Hw

validatorr:


RomanYuzviak (commented):
Great solution!

RomanYuzviak (approved):

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,23 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final String NATIONALITY = "Ukrainian";
+    private static final String REGEX = "-";
```
NazarSmith228, line 8:
```suggestion
    private static final String LIVING_PERIOD_SEPARATOR= "-";
```

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,23 @@
...
+    private static final int REQUIRED_AGE = 35;
+    private static final int AGE_TO_LIVE_IN_UKRAINE = 10;
```
NazarSmith228, line 10:
```suggestion
    private static final int YEARS_TO_LIVE_IN_UKRAINE = 10;
```

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,23 @@
...
+    private static final int STARTED_TO_LIVE = 0;
+    private static final int ENDED_TO_LIVE = 1;
```
NazarSmith228, line 12:
```suggestion
    private static final int LIVING_PERIOD_START_INDEX = 0;
    private static final int LIVING_PERIOD_END_INDEX = 1;
```

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,23 @@
...
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] periodsInUkraine = candidate.getPeriodsInUkr().split(REGEX);
+        int firstPart = Integer.parseInt(periodsInUkraine[STARTED_TO_LIVE]);
+        int secondPart = Integer.parseInt(periodsInUkraine[ENDED_TO_LIVE]);
```
NazarSmith228, line 18:
```suggestion
        int livingPeriodStart= Integer.parseInt(periodsInUkraine[STARTED_TO_LIVE]);
        int livingPeriodEnd= Integer.parseInt(periodsInUkraine[ENDED_TO_LIVE]);
```

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,23 @@
...
+        return candidate.getAge() >= REQUIRED_AGE && candidate.isAllowedToVote()
+                && candidate.getNationality().equals(NATIONALITY)
+                && (secondPart - firstPart) > AGE_TO_LIVE_IN_UKRAINE;
```
NazarSmith228, line 21: you can create a separate variable for substraction in order to improve readability

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +17,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(number -> number.split(","))
```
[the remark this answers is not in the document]
NazarSmith228: it would be better to name it `s` - short names are better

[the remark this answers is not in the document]
NazarSmith228: agree, we can use single `flatMap` instead of `map` + `flatMap`

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +32,12 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .mapToDouble(element -> element % 2 != 0
+                        ? numbers.get(element) - 1 : numbers.get(element))
```
NazarSmith228, line 37: why to Double?

also it is always better to avoid negation when possible (you can use `==` instead of `!=` and swap)

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +32,12 @@ public int findMinEvenNumber(List<String> numbers) {
...
+                .filter(element -> element % 2 != 0)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
NazarSmith228, line 40: `NoSuchElementException` is thrown by default here - you can omit the supplier

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +49,10 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getAge() >= fromAge && person.getAge() <= toAge
+                        && person.getSex() == Person.Sex.MAN)
```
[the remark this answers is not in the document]
NazarSmith228: you can create a separate `Predicate` variable and use it here instead of inlined

you can also make this predicate constant

if you wish to inline - then indeed, checking the sex first might be better for performance

NazarSmith228 (changes requested):

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,24 @@
...
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] periodsInUkraine = candidate.getPeriodsInUkr().split(LIVING_PERIOD_SEPARATOR);
+        int livingPeriodStart = Integer.parseInt(periodsInUkraine[LIVING_PERIOD_START_INDEX]);
+        int livingPeriodEnd = Integer.parseInt(periodsInUkraine[LIVING_PERIOD_END_INDEX]);
+        int resultOfSubtraction = livingPeriodEnd - livingPeriodStart;
+        return candidate.getAge() >= REQUIRED_AGE && candidate.isAllowedToVote()
+                && candidate.getNationality().equals(NATIONALITY)
+                && (resultOfSubtraction) > YEARS_TO_LIVE_IN_UKRAINE;
```
NazarSmith228, line 22:
```suggestion
        int yearsInUkraine = livingPeriodEnd - livingPeriodStart;
        return candidate.getAge() >= REQUIRED_AGE && candidate.isAllowedToVote()
                && candidate.getNationality().equals(NATIONALITY)
                && yearsInUkraine > YEARS_TO_LIVE_IN_UKRAINE;
```

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +30,12 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .mapToDouble(element -> element % 2 == 1
```
NazarSmith228, line 34: I still don't understand why you're mapping to Double here :)

NazarSmith228 (changes requested):
One last step :)

src/main/java/practice/StreamPractice.java
validatorr: My bad, forgot to change it to map.

NazarSmith228 (approved):
GTG

PR #877, opened 2023-09-12
Stream practice implementation

KDanil2000:


src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +80,11 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex() == Person.Sex.WOMAN && p.getAge() >= femaleAge)
```
SalnikovMarkiian, line 84: I have an interesting question: what if the list of cats is empty? 
in my impl i checked additionally `  !p.getCats.isEmpty()`
it`s optioanl for u

src/main/java/practice/StreamPractice.java
```diff
@@ -1,8 +1,11 @@
...
-import java.util.Collections;
+import java.util.Arrays;
 import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import model.Candidate;
+import model.Cat;
 import model.Person;
 
 public class StreamPractice {
```
SalnikovMarkiian, line 11: it would be better to remove extra comments in finish code

SalnikovMarkiian (approved):
Cool !) I think this is a good solution, but I will add a few optional changes in comments

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +17,14 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(i -> i.split(","))
+                .flatMap(Arrays::stream)
```
Andriy-dz, line 22: you can combine these methods

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +33,11 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(i -> i % 2 == 1 ? numbers.get(i) - 1 : numbers.get(i))
+                .filter(n -> n % 2 == 1)
+                .average()
+                .orElseThrow();
```
Andriy-dz, line 40: what to throw then?

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +49,10 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                    .filter(person -> person.getSex() == Person.Sex.MAN)
+                    .filter(person -> person.getAge() >= fromAge && person.getAge() <= toAge)
```
Andriy-dz, line 54: you can combine these methods

Andriy-dz (changes requested):
Good job! Let`s improve your solution)

Elena-Bruyako (approved):

PR #880, opened 2023-09-13
imp metods

Petro-Smoliar:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +21,15 @@ public class StreamPractice {
...
-        return 0;
+        int result = numbers.stream()
+                .map(n -> n.split(DELIMITER))
+                .flatMap(Arrays::stream)
+                .map(Integer::parseInt)
+                .reduce(Integer.MAX_VALUE, (a,b) -> b < a && pairingCheck(b) ? b : a);
+        if (result == Integer.MAX_VALUE) {
+            throw new RuntimeException("Can't get min value from list: " + numbers);
+        }
+        return result;
```
olekskov, line 32: Most obvious and readable solution is better in most cases
you reduce command is far from easy to be understood
```suggestion
        return result = numbers.stream()
                .map(numbersString -> numbersString.split(DELIMITER))
                .flatMap(Arrays::stream)
                .map(Integer::parseInt)
                .filter(i -> i % 2 == 0)
                .min()
                .orElseThrow(() -> new RuntimeException("Can't get min value from list: " + numbers));
```

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +38,14 @@ public int findMinEvenNumber(List<String> numbers) {
...
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .mapToObj(index -> !pairingCheck(index) ? numbers.get(index) - SUBTRACTION_VALUE
+                        : numbers.get(index))
+                .mapToInt(number -> number)
+                .filter(number -> !pairingCheck(number))
+                .average()
+                .orElseThrow(() -> new NoSuchElementException(
+                        "The average of odd numbers cannot be calculated"));
     }
```
olekskov, line 49: 1. Boolean method are more understandable when named with 'is'. Like isEvenNumber, isValidRecord etc.
2. Ternaries are not easiest structure for reading so should be properly formatted
3. exclamation mark (!) in code is very easy to be missed. When not a problem it's better to create opposite method not to use !
4. Why mapToObj to return to int?
```suggestion
                .map(index -> isOddNumber(index) 
                        ? numbers.get(index) - SUBTRACTION_VALUE
                        : numbers.get(index))
                .filter(number -> isOddNumber(number))
                .average()
                .orElseThrow(() -> new NoSuchElementException(
                        "The average of odd numbers cannot be calculated"));
    }
```

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +57,11 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex() == Person.Sex.MAN
+                        && p.getAge() >= fromAge
+                        && p.getAge() <= toAge)
+                .collect(Collectors.toList());
```
olekskov, line 64: Never use 1-2 letter abbreviations for naming, even in steams and lambdas.
In a week when your forget code context you will hate such decision.
Please recheck all methods
```suggestion
        return peopleList.stream()
                .filter(person -> person.getSex() == Person.Sex.MAN
                        && person.getAge() >= fromAge
                        && person.getAge() <= toAge)
                .collect(Collectors.toList());
```

olekskov (changes requested):

Ivan95kos (approved):
Nice

PR #891, opened 2023-09-27
added solution

andriihomenko:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,26 @@
...
 
-public class CandidateValidator {
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
     //write your code here
+    private static final int REQUIRED_AGE = 35;
+    private static final String REQUIRED_NATIONALITY = "Ukrainian";
+    private static final int REQUIRED_YEARSINUKRAINE = 10;
```
Rommelua, line 10:
```suggestion
    private static final int REQUIRED_YEARS_IN_UKRAINE = 10;
```

src/main/java/practice/StreamPractice.java
```diff
@@ -14,16 +21,29 @@ public class StreamPractice {
...
+         */
+        return numbers.stream()
+                .flatMap(n -> Arrays.stream(n.split(",")))
+                .map(Integer::valueOf)
+                .filter(num -> num % 2 == 0)
+                .sorted()
+                .findFirst()
+                .orElseThrow(()
+                        -> new RuntimeException("Can't get min value from list: method_input_list: "
+                        + numbers));
```
Rommelua, line 37: image

src/main/java/practice/StreamPractice.java
```diff
@@ -50,16 +73,29 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        Predicate<Person> pred = p -> p.getSex().equals(Person.Sex.MAN);
```
Rommelua, line 76: image

Rommelua (changes requested):
Before you create PR with your homework, you need to go through the checklist under the task and correct all the points described there. The mentor will not check the work until the checklist points are corrected

src/main/java/practice/StreamPractice.java
andriihomenko: We can use both == and equals(). I think there are more arguments for using .equals() for enum types.

Rommelua (approved):

PR #893, opened 2023-09-29
solved task "StreamPractice" in module "javaCore/Java8-Part2"

Krasnov-Maksim:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,18 @@
...
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int AGE_THRESHOLD = 35;
+    private static final int PERIOD_THRESHOLD = 10;
+
+    @Override
+    public boolean test(final Candidate candidate) {
+        int startYear = Integer.parseInt(candidate.getPeriodsInUkr().split("-")[0]);
```
sokimaaa, line 12: candidate.getPeriodsInUkr().split("-") duplicates.
should be extracted into local var.

sokimaaa, line 12: all indexes should be constant.

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(word -> word.split(","))
+                .flatMap(Arrays::stream)
```
sokimaaa, line 23: use one mapping whenever its possible 
```suggestion
                .flatMap(str -> Arrays.stream(str.split(",")))
```

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +33,16 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        int[] positionHolder = new int[1];
+        return numbers.stream()
+                .mapToInt(number -> {
+                    final int checkedNumber = positionHolder[0] % 2 != 0 ? number - 1 : number;
+                    positionHolder[0]++;
+                    return checkedNumber;
```
sokimaaa, line 41: changing outside variables in stream is not recommended and may cause the `side effect`. 

think about using IntStream.range(...) to avoid using `positionHolder`

sokimaaa (changes requested):
see comment above

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,21 @@
...
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int AGE_THRESHOLD = 35;
+    private static final int PERIOD_THRESHOLD = 10;
+    private static final int INDEX_0 = 0;
+    private static final int INDEX_1 = 1;
```
sokimaaa, line 10: constant name is confusing, make it self-descriptive.

sokimaaa (changes requested):

aaananas (approved):

PR #895, opened 2023-10-04
StreamPractice of Starikova Olga

OlgaStarikova:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,28 @@
...
+        String[] years = candidate.getPeriodsInUkr().split(YEARS_REGEX_SPLIT);
+        int yearsCount = Integer.parseInt(years[YEARS_INDEX_END])
+                - Integer.parseInt(years[YEARS_INDEX_START]);
+        if (candidate.isAllowedToVote()
+                && candidate.getAge() >= AGE_MIN_VALID
+                && candidate.getNationality().equals(NATIONALITY_VALID)
+                && yearsCount >= YEARS_MIN_VALID) {
+            return true;
+        }
+        return false;
```
okuzan, line 26: Consider simplifying the code. You can directly return the result of the condition instead of using if/else or ternary operation to return true/false. For example:

// Instead of this
if (condition) {
return true;
} else {
return false;
}

// You can do this
return condition;
This makes the code cleaner and more straightforward.

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,28 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+
```
```suggestion
```

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,28 @@
...
+    private static final int AGE_MIN_VALID = 35;
+    private static final String NATIONALITY_VALID = "Ukrainian";
```
okuzan, line 9:
```suggestion
    private static final String VALID_NATIONALITY = "Ukrainian";
```

okuzan (changes requested):
Don't use one-letter naming in streams outside of Math context, e.g. change `p` to `person`

olekskov (approved):

PR #896, opened 2023-10-06
Added solution

slaybrute:


okuzan (changes requested):
replace runtime with nosuchelement, extract all literals from validator class to constants, don't use if-else in the stream chains

Ivan95kos (approved):
Great job

PR #898, opened 2023-10-09
Complete all Stream Api tasks

Fantom5511:


sokimaaa (changes requested):

sokimaaa (changes requested):
see comment below

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,29 @@
 package practice;
 
-public class CandidateValidator {
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MIN_AGE = 35;
```
kerrrusha, line 7: it's better to put fields in order by their types for better code readability, like:
int a...
int b...
String c...

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,29 @@
...
+        int yearsInUkraine = yearTo - yearFrom;
+
+        return candidate.getAge() >= MIN_AGE
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals(NATIONALITY)
+                && yearsInUkraine >= YEARS_IN_UKRAINE;
+    }
+
+
     //write your code here
```
kerrrusha, line 28: would be nice to delete these blank lines

kerrrusha (changes requested):
need cosmetic fixes, but overall it's nice

kerrrusha (approved):

PR #908, opened 2023-10-19
Practiced in Streams

HovorukhaBohdan:


src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +43,25 @@ public int findMinEvenNumber(List<String> numbers) {
...
+        long numOfOddElements = modifiedNumbers.stream()
+                .filter(n -> n % 2 != 0)
+                .mapToInt(n -> n)
+                .count();
+
+        if (numOfOddElements == 0) {
+            throw new NoSuchElementException("Unexpected numbers");
+        }
+
+        return (double) sumOfOddElements / numOfOddElements;
```
dima-semeniuk, line 64: Your code looks very complicated. You can use ' IntStream.range ' method to simplify this implementation. For example: IntStream.range(0, numbers.size())
  and then apply the filter

dima-semeniuk (approved):

src/main/java/practice/StreamPractice.java
```diff
@@ -1,20 +1,37 @@
...
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
+
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(i -> i.split(SPLIT_CHARACTER))
+                .map(s -> Arrays.stream(s)
+                        .mapToInt(Integer::parseInt)
+                        .min()
```
kshuryhin, line 29: I would say we should not use `min()` here, as we will do that at the end of stream chaining

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +40,21 @@ public int findMinEvenNumber(List<String> numbers) {
...
+        long numOfOddElements = IntStream.range(0, numbers.size())
+                .map(i -> i % 2 == 0 ? numbers.get(i) : numbers.get(i) - 1)
+                .filter(n -> n % 2 != 0)
+                .count();
+
+        if (numOfOddElements == 0) {
+            throw new NoSuchElementException("Unexpected numbers");
+        }
+
+        return (double) sumOfOddElements / numOfOddElements;
```
kshuryhin, line 57: Cool, but try to go with more accurate approach. Recommend to take a look at this class 
https://docs.oracle.com/javase/8/docs/api/java/util/LongSummaryStatistics.html

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +84,14 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        Predicate<Person> testPerson = p -> (p.getSex() == Person.Sex.MAN
```
kshuryhin, line 87: Better name for the predicate should be provided

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +84,14 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
...
+                && p.getAge() >= fromAge && p.getAge() <= maleToAge)
+                || (p.getSex() == Person.Sex.WOMAN
+                && p.getAge() >= fromAge && p.getAge() <= femaleToAge);
+
+        return peopleList.stream()
+                .filter(testPerson)
+                .collect(Collectors.toList());
```
kshuryhin, line 94:
```suggestion
                .toList();
```

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +125,11 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        CandidateValidator test = new CandidateValidator();
```
kshuryhin, line 128: Better name for the validator

kshuryhin (changes requested):
Nice try, but some changes should be done

kshuryhin (approved):
Great job!

PR #909, opened 2023-10-23
first_commit

Lupen07:


okuzan (changes requested):
don't use short naming inside of stream chains, for example change p to person. Make string literals like "," '-" - constants too

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +32,12 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(i -> (i % 2 == 1) ? numbers.get(i) - 1 : numbers.get(i))
```
kozhukhovsky, line 36:
```suggestion
                .map(index -> (i % 2 == 1) ? numbers.get(i) - 1 : numbers.get(i))
```

kozhukhovsky (approved):

PR #911, opened 2023-10-26
hw-stream-api

IhorTrokhymchuk:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,20 @@
...
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int PERIOD_TO_NUMBER = 1;
+    private static final int PERIOD_FROM_NUMBER = 0;
+    private static final int PERIOD_IN_UKRAINE = 10;
+
+    @Override
+    public boolean test(Candidate c) {
```
okuzan, line 12:
```suggestion
    public boolean test(Candidate candidate) {
```
always go for readability - we, as devs, more often read code, than write it

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,20 @@
...
+        return c.getAge() >= 35
+                && c.isAllowedToVote()
+                && "Ukrainian".equals(c.getNationality())
```
okuzan, line 15: 35 and Ukrainian should be constants

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,20 @@
...
+                && Integer.parseInt(c.getPeriodsInUkr().split("-")[PERIOD_TO_NUMBER])
```
okuzan, line 16: - should be constant - like SEPARATOR/DELIMITER

okuzan (changes requested):
rename s n i p c in stream chains for better readability

kozhukhovsky (approved):

PR #912, opened 2023-10-26
Created first version of the program.

ponomvrenko:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,27 @@
...
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MIN_YEARS_OLD = 35;
+    private static final int MIN_TOTAL_DURATION = 10;
+    private static final String VALID_NATIONALITY = "Ukrainian";
+    private static final String REGEX = "-";
+    private static final int SINCE_YEAR_INDEX = 0;
+    private static final int TO_YEAR_INDEX = 1;
+
+    @Override
+    public boolean test(Candidate c) {
```
proxychang, line 15: I would recommend to change Candidate variable name to something more informational

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +17,15 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        if (numbers.size() == 0) {
+            throw new RuntimeException("Can't get min value from list:" + numbers);
+        }
```
proxychang, line 22: It will mcuh better to delete this if/else construction and use one of [Optional ](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html) methods

proxychang: In addition, you can make exception message constant

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +17,15 @@ public class StreamPractice {
...
+        return numbers.stream()
+                .map(s -> List.of(s.replace(" ", "").split(",")))
```
proxychang, line 24: You can make regex constant. And why do you need s.replace here? It seems we can avoid using replace method

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +17,15 @@ public class StreamPractice {
...
+                .flatMapToInt(l -> l.stream().mapToInt(Integer::parseInt))
```
proxychang, line 25: Yes, it works properly. But it can be simplified by using just flatmap. You don`t need to change it, but think about it.

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +52,10 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> fromAge <= p.getAge() && p.getAge() <= toAge
+                        && p.getSex() == Person.Sex.MAN)
```
proxychang, line 57: It`ll better to compare getSex and Sex.VALUE by equals

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +84,12 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex() == Person.Sex.WOMAN && p.getAge() >= femaleAge)
+                .map(Person::getCats)
+                .flatMap(Collection::stream)
+                .map(Cat::getName)
+                .collect(Collectors.toList());
```
proxychang, line 92: Maybe you can simplify it by using .toList() instead of .collect(Collectors.toList())

proxychang (changes requested):
You`ve done well-wrote solution! I commented few points that will do your code a little optimized, pay attention to it please. And do not forget magical numbers and constant fields.

src/main/java/practice/StreamPractice.java
ponomvrenko: Great suggestion, it removes useless if-else with orElseThrow() method

src/main/java/practice/StreamPractice.java
ponomvrenko: Haha, honestly idk, I just imagine that we have spaces around our numbers, and we have to handle this situation. Thanks!

src/main/java/practice/StreamPractice.java
ponomvrenko: It is permissible to use the == operator in place of the equals method when comparing two object references if it is known that at least one of them refers to an enum constant.

https://stackoverflow.com/questions/1750435/comparing-java-enum-members-or-equals/1750453#1750453

ponomvrenko:
@proxychang thanks for your review! I just change my code with your suggestions, and it becomes more elegant <3

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(s -> List.of(s.split(REGEX)))
+                .flatMap(Collection::stream)
+                .mapToInt(Integer::parseInt)
+                .filter(n -> n % 2 == 0)
+                .min()
```
mnyshenko, line 28: It's better to give more readable names for your parameters in lambdas. One-letter names are usually used only in for-loop declarations

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +35,13 @@ public int findMinEvenNumber(List<String> numbers) {
...
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(i -> i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i))
+                .boxed()
+                .filter(n -> n % 2 != 0)
+                .mapToDouble(n -> n)
+                .average()
+                .getAsDouble();
     }
```
mnyshenko, line 45: A few points here:
1) Naming, one-letter naming are acceptable only when creating for-loop
2) a few redundant actions here; please take a look at given implementation
3) don't repeat yourself and extract identical parts inside methods/variables for further reuse: n -> n % 2 != 0

```
public static Double getOddNumsAverage(List<Integer> numbers) {
                .map(index -> isOddNumber(index) ? numbers.get(index) - 1 : numbers.get(index))
                .filter(NextBiggerNumber::isOddNumber)
                .average()
                .getAsDouble();
    }

    private static boolean isOddNumber(int number) {
        return number % 2 != 0;
    }
```

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +53,10 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> fromAge <= p.getAge() && p.getAge() <= toAge
+                        && p.getSex() == Person.Sex.MAN)
```
mnyshenko, line 58: One-letter naming

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +71,12 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> (p.getSex() == Person.Sex.MAN && p.getAge() >= fromAge
+                        && p.getAge() <= maleToAge)
+                        || (p.getSex() == Person.Sex.WOMAN && p.getAge() >= fromAge
+                        && p.getAge() <= femaleToAge))
```
mnyshenko, line 78: p.getAge() used twice so can be extracted to a separate variable, same with sex

mnyshenko (changes requested):
A few point to consider

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,27 @@
...
+    private static final int MIN_TOTAL_DURATION = 10;
+    private static final String VALID_NATIONALITY = "Ukrainian";
+    private static final String REGEX = "-";
+    private static final int SINCE_YEAR_INDEX = 0;
+    private static final int TO_YEAR_INDEX = 1;
+
+    @Override
+    public boolean test(Candidate candidate) {
+        if (candidate.getAge() >= MIN_YEARS_OLD && candidate.isAllowedToVote()
+                && candidate.getNationality().equals(VALID_NATIONALITY)) {
```
bhdnchui, line 17:
```suggestion
                && VALID_NATIONALITY.equals(candidate.getNationality())) {
```
In that way, you will be sure that you never will get NPE.

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +36,15 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(index -> isOddNumber(index) ? numbers.get(index) - 1 : numbers.get(index))
+                .filter(StreamPractice::isOddNumber)
+                .average()
```
bhdnchui, line 42: Check if there is an average number. If there is not, throw NoSuchElementException according to the task requirements.

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +74,17 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
...
+            int age = person.getAge();
+            Person.Sex sex = person.getSex();
+
+            return (sex == Person.Sex.MAN && age >= fromAge && age <= maleToAge)
+                    || (sex == Person.Sex.WOMAN && age >= fromAge && age <= femaleToAge);
+        };
+
+        return peopleList.stream()
+                .filter(isAgeInRange)
+                .collect(Collectors.toList());
```
bhdnchui, line 87:
```suggestion
                .toList();
```

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +93,12 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex() == Person.Sex.WOMAN && p.getAge() >= femaleAge)
+                .map(Person::getCats)
+                .flatMap(Collection::stream)
+                .map(Cat::getName)
+                .collect(Collectors.toList());
```
bhdnchui, line 101:
```suggestion
                .toList();
```

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +114,11 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        CandidateValidator candidateValidator = new CandidateValidator();
+        return candidates.stream()
+                .filter(candidateValidator)
+                .map(Candidate::getName)
+                .sorted()
+                .collect(Collectors.toList());
```
bhdnchui, line 122:
```suggestion
                .toList();
```

bhdnchui (changes requested):
Good job, but a few minor points need to consider.

src/main/java/practice/StreamPractice.java
ponomvrenko: @bhdnchui I have there getAsDouble() method that includes 
`if (!isPresent) {
            throw new NoSuchElementException("No value present");
}`
Is my solution here incorrect?

bhdnchui (approved):
Good job!

PR #930, opened 2023-10-30
implemented all the requested methods

annton-shakur:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,16 @@
...
+
+public class CandidateValidator implements Predicate<Candidate> {
+
+    @Override
+    public boolean test(Candidate candidate) {
+        return (candidate.getAge() >= 35
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals("Ukrainian")
+                && Integer.parseInt(candidate.getPeriodsInUkr().split("-")[1])
+                - Integer.parseInt(candidate.getPeriodsInUkr().split("-")[0]) >= 10);
```
DEz-1996, line 14: Create constants for magic values
Also, optionaly, you can remove the logic for calculating years of life in a country into a separate method

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,13 @@ public class StreamPractice {
...
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(string -> Arrays.stream(string.split(","))
+                        .map(Integer::parseInt))
+                .filter(integer -> integer % 2 == 0)
+                .min(Comparator.comparing(integer -> integer))
+                .orElseThrow(() -> new RuntimeException("Can't get min value from list:"
+                        + " < Here is our input 'numbers' >"));
```
DEz-1996, line 28: You can outsource creating exception in the Supplier and put them in the .orElseThrow()
This will improve the readability of the code

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +68,21 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
...
+                        && person.getAge() >= fromAge && person.getAge() <= maleToAge)
+                        || (person.getSex().equals(Person.Sex.WOMAN)
+                        && person.getAge() >= fromAge && person.getAge() <= femaleToAge)) {
+                    return true;
+                }
+                return false;
+            }
+        };
+        return peopleList.stream()
+                .filter(predicate)
```
DEz-1996, line 84: It seems that it is better to think about naming of "predicate"
By what principle does it filter people?

DEz-1996 (changes requested):
Good job, but please pay attention to my comments

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,21 @@
...
+
+    @Override
+    public boolean test(Candidate candidate) {
+        return (candidate.getAge() >= CANDIDATE_MINIMUM_AGE
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals(UKRAINIAN_NATIONALITY)
+                && Integer.parseInt(candidate.getPeriodsInUkr().split(HYPHEN_FOR_SPLIT)[1])
+                - Integer.parseInt(candidate.getPeriodsInUkr().split(HYPHEN_FOR_SPLIT)[0])
+                >= MIN_TIME_LIVED_IN_UKRAINE);
+    }
```
mnyshenko, line 20: 1. Outer parentheses can be omitted
2. candidate.getPeriodsInUkr().split() is used twice here, which has a bad impact on performance, You can store the result of the split in a separate variable and reuse it

src/main/java/practice/StreamPractice.java
```diff
@@ -1,20 +1,34 @@
...
+
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(string -> Arrays.stream(string.split(COMMA))
+                        .map(Integer::parseInt))
+                .filter(integer -> integer % 2 == 0)
+                .min(Comparator.comparing(integer -> integer))
+                .orElseThrow(() -> new RuntimeException(NO_MIN_VALUE_MESSAGE
+                        + numbers));
```
mnyshenko, line 31: let's rewrite this stream
1) Let's use mapToInt after flatMap instead of using .map inside of flatMap; this will help to omit using Comparator.comparing(integer -> integer) in min
2) integer -> integer % 2 == 0, this statement is written multiple times in this class, let's make a separate method for executing this logic and reuse it everywhere

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +37,11 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(index -> index % 2 != 0 ? numbers.get(index) - 1 : numbers.get(index))
+                .filter(number -> number % 2 != 0)
+                .average()
+                .orElseThrow();
```
mnyshenko, line 44: index % 2 != 0 duplication!

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +53,10 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getSex().equals(Person.Sex.MAN))
```
mnyshenko, line 57: https://github.com/mate-academy/jv-stream-github-practice/blob/main/checklist.md#pay-attention-to-what-is-a-better-way-to-compare-enum-values-equals-vs--

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +108,10 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(candidate -> new CandidateValidator().test(candidate))
```
mnyshenko, line 112: we should not create CandidateValidator instance each time we call validateCandidates method, this predicate can be instantiated only once for better performance

mnyshenko (changes requested):
Great, but there are a few point to consider

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +71,13 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        Predicate<Person> workingAgePredicate = person -> (Person.Sex.MAN.equals(person.getSex())
+                && person.getAge() >= fromAge && person.getAge() <= maleToAge)
+                || (Person.Sex.WOMAN.equals(person.getSex())
```
mnyshenko, line 76: https://github.com/mate-academy/jv-stream-github-practice/blob/main/checklist.md#pay-attention-to-what-is-a-better-way-to-compare-enum-values-equals-vs--

don't forget how to better compare enums

mnyshenko (approved):
Approved, but pay attention to the comment regarding enum comparison

PR #932, opened 2023-10-31
Realization of StreamPractice

forzen123212:


src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +55,11 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex() == Person.Sex.MAN)
+                .filter(p -> p.getAge() >= fromAge
```
kshuryhin, line 60: Try not to use several `filter` operations here, but rather create predicate/method for that

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +92,13 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex() == Person.Sex.WOMAN)
+                .filter(p -> p.getCats() != null)
+                .filter(p -> p.getAge() >= femaleAge)
```
kshuryhin, line 98: Same here, try to leverage separate predicate/method approach

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +92,13 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
...
+                .flatMap(person -> person.getCats().stream())
+                .map(Cat::getName)
+                .collect(Collectors.toList());
```
kshuryhin, line 101: JFYI you can simply use `toList()` now in Java.

kshuryhin (changes requested):
Great job! Just some points to consider

src/main/java/practice/StreamPractice.java
kshuryhin: Yes, but on the other hand it brings worse performance. So I think we can find an approach both readable enough and with solid performance

kshuryhin (approved):

PR #937, opened 2023-11-07
implemented stream practice methods

LaverAndrii:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,14 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(n -> n.split(","))
+                .flatMap(Arrays::stream)
```
kerrrusha, line 25: could be done only via `flatMap`

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,14 @@ public class StreamPractice {
...
+                .map(Integer::parseInt)
+                .filter(n -> n % 2 == 0)
+                .min(Integer::compareTo)
```
kerrrusha, line 28: `Integer::compareTo` - is it neccessary to explicit?

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +36,12 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        AtomicInteger index = new AtomicInteger();
+        return numbers.stream()
+                .mapToInt(n -> index.getAndIncrement() % 2 != 0 ? n - 1 : n)
```
kerrrusha, line 41: use `IntStream.range(0, numbers.size())` to avoid using AtomicInteger

kerrrusha (changes requested):

Rommelua (approved):

PR #942, opened 2023-11-13
Implemented methods of StreamPractice.class and implemented Predicate…

Sanmartial:
… for CandidateValidator.class

### DIFF
--- a/src/main/java/practice/CandidateValidator.java
+++ b/src/main/java/practice/CandidateValidator.java
@@ -1,5 +1,28 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final String NATIONALITY = "Ukrainian";
+    private static final int MIN_AGE_FOR_ELECTION = 35;
+    private static final int PERIOD_IN_COUNTRY = 10;
+    private static final String DELIMITER_DASH = "-";
+
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= MIN_AGE_FOR_ELECTION
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals(NATIONALITY)
+                && periodOfStay(candidate.getPeriodsInUkr()) >= PERIOD_IN_COUNTRY;
+    }
+
+    private int periodOfStay(String periodsInUkr) {
+        if (periodsInUkr == null) {
+            return 0;
+        }
+        String[] years = periodsInUkr.split(DELIMITER_DASH);
+        return Integer.parseInt(years[1]) - Integer.parseInt(years[0]);
+    }
+
 }
--- a/src/main/java/practice/StreamPractice.java
+++ b/src/main/java/practice/StreamPractice.java
@@ -1,11 +1,26 @@
 package practice;
 
-import java.util.Collections;
+import java.util.Arrays;
+import java.util.Comparator;
 import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import model.Candidate;
+import model.Cat;
 import model.Person;
 
 public class StreamPractice {
+
+    private static final String DELIMITER_COMMA = ",";
+
+    private final CandidateValidator validator;
+
+    public StreamPractice() {
+        validator = new CandidateValidator();
+    }
+
     /**
      * Given list of strings where each element contains 1+ numbers:
      * input = {"5,30,100", "0,22,7", ...}
@@ -14,7 +29,14 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+
+        return numbers.stream()
+                .map(str -> str.split(DELIMITER_COMMA))
+                .flatMap(str -> Arrays.stream(str).map(Integer::parseInt))
+                .filter(number -> (number % 2) == 0)
+                .mapToInt(n -> n)
+                .min().orElseThrow(() ->
+                        new RuntimeException("Can't get min value from list: method_input_list"));
     }
 
     /**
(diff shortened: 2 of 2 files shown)


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,28 @@
...
+                && candidate.getNationality().equals(NATIONALITY)
+                && periodOfStay(candidate.getPeriodsInUkr()) >= PERIOD_IN_COUNTRY;
+    }
+
+    private int periodOfStay(String periodsInUkr) {
+        if (periodsInUkr == null) {
+            return 0;
+        }
+        String[] years = periodsInUkr.split(DELIMITER_DASH);
+        return Integer.parseInt(years[1]) - Integer.parseInt(years[0]);
```
sokimaaa, line 25: Indexes should be constant with proper name

src/main/java/practice/StreamPractice.java
```diff
@@ -1,11 +1,26 @@
...
 
 public class StreamPractice {
+
+    private static final String DELIMITER_COMMA = ",";
+
+    private final CandidateValidator validator;
+
+    public StreamPractice() {
+        validator = new CandidateValidator();
+    }
```
sokimaaa, line 22: its needed for one method only, you can instantiate there
```suggestion
```

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +29,14 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+
+        return numbers.stream()
+                .map(str -> str.split(DELIMITER_COMMA))
+                .flatMap(str -> Arrays.stream(str).map(Integer::parseInt))
```
sokimaaa, line 35: try to reduce map operations as much as possible

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +45,16 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+
+        List<Integer> modifiedList = IntStream.range(0, numbers.size())
+                .mapToObj(i -> i % 2 == 1 ? numbers.get(i) - 1 : numbers.get(i))
+                .collect(Collectors.toList());
```
sokimaaa, line 51: YOu can proceed stream opetions. 
dont need to collect here.
```suggestion
                .collect(Collectors.toList());
```

sokimaaa (changes requested):

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +39,14 @@ public int findMinEvenNumber(List<String> numbers) {
...
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+
+        return IntStream.range(0, numbers.size())
+                .mapToObj(i -> i % 2 == 1 ? numbers.get(i) - 1 : numbers.get(i))
+                .toList().stream()
+                .filter(num -> num % 2 != 0)
+                .mapToDouble(Integer::doubleValue)
+                .average()
```
sokimaaa, line 48: Reduce the number of map operations as much as possible.
Check other methods.

see example:
```suggestion
                .mapToDouble(i -> i % 2 == 1 ? numbers.get(i) - 1 : numbers.get(i))
                .toList().stream()
                .filter(num -> num % 2 != 0)
                .average()
```

sokimaaa (changes requested):

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +23,14 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+
+        return numbers.stream()
+                .flatMap(str -> Arrays.stream(str.split(DELIMITER_COMMA)))
+                .map(Integer::parseInt)
```
Sanderfuck, line 29: Also you can use MapToInt method

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +56,12 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+
+        return peopleList.stream()
+                .filter(person -> person.getSex() == Person.Sex.MAN
+                        && person.getAge() >= fromAge
+                        && person.getAge() <= toAge)
```
Sanderfuck, line 63: Here you can use Predicate like in getWorkablePeople method, it increase readability

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +76,16 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        Predicate<Person> people = (person) -> {
+            if (person.getSex() == Person.Sex.MAN) {
+                return person.getAge() >= fromAge && person.getAge() <= maleToAge;
+            } else {
+                return person.getAge() >= fromAge && person.getAge() <= femaleToAge;
+            }
```
Sanderfuck, line 84: Can you use here this construction? 
image

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +116,13 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        CandidateValidator validator = new CandidateValidator();
+
+        return candidates.stream()
+                .filter(validator::test)
```
Sanderfuck, line 122: Here you can use only object, because new CandidateValidator() implement predicate 
image

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +116,13 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
...
+                .sorted(Comparator.comparing(Candidate::getName)
+                        .thenComparingInt(Candidate::getAge))
+                .map(Candidate::getName)
```
Sanderfuck, line 125: At first you sorted, what if we map at first and after make sorting, do we need any comparator in this method? Let`s check!

Sanderfuck (changes requested):

kerrrusha (approved):

PR #944, opened 2023-11-14
practiced stream API, did CandidateValidator class

maxymmusiienko:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,13 @@ public class StreamPractice {
...
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(str -> Arrays.stream(str.split(",")))
+                .mapToInt(Integer::parseInt)
+                .filter(n -> n % 2 == 0)
+                .min()
+                .orElseThrow(() ->
+                        new RuntimeException("Can't get min value from list: " + numbers));
```
okuzan, line 28:
```suggestion
                        new NoSuchElementException("Can't get min value from list: " + numbers));
```

okuzan (changes requested):

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +34,13 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .mapToObj(i -> i % 2 == 0 ? numbers.get(i) : numbers.get(i) - 1)
```
Sanderfuck, line 38: You can use just map

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +34,13 @@ public int findMinEvenNumber(List<String> numbers) {
...
+                .filter(n -> n % 2 != 0)
+                .mapToDouble(Double::valueOf)
```
Sanderfuck, line 40: You don`t need it

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +86,13 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getSex().equals(Person.Sex.WOMAN)
+                        && person.getAge() >= femaleAge)
+                .map(Person::getCats)
+                .flatMap(Collection::stream)
+                .map(Cat::getName)
```
Sanderfuck, line 94: you don`t need 2 methods map

Sanderfuck (changes requested):

Sanderfuck (changes requested):
Check comments

kozhukhovsky (approved):

PR #945, opened 2023-11-15
Implemented StreamPractice methods, CandidateValidator as Predicate

promahos:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,14 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        OptionalInt minValue = numbers.stream()
```
Rommelua, line 23:
```suggestion
        return numbers.stream()
```

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,14 @@ public class StreamPractice {
...
+                .map(s -> s.split(","))
+                .flatMap(Arrays::stream)
```
Rommelua, line 25: combine to one flatMap call

Rommelua (changes requested):
good job

Rommelua (approved):

PR #948, opened 2023-11-18
jv-stream-solution

Valdemar3X:


src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +37,12 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .mapToObj(i -> i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i))
+                .filter(n -> n % 2 != 0)
```
olekskov, line 42: Please don't use 1-2 letter abbreviations even in streams, they became unreadable in 2 days after you wrote them
```suggestion
                .filter(number -> number % 2 != 0)
```
Please recheck whole code for same issue

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +21,14 @@ public class StreamPractice {
...
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        Optional<Integer> min = numbers.stream()
+                .map(s -> s.split(","))
+                .flatMap(a -> Arrays.stream(a))
+                .map(s -> Integer.parseInt(s))
+                .filter(integer -> integer % 2 == 0 && integer != 1)
+                .min(Comparator.naturalOrder());
+        return min.orElseThrow(() ->
+                new RuntimeException("Can't get min value from list:" + numbers));
```
olekskov, line 31:
```suggestion
        return numbers.stream()
                .map(s -> s.split(","))
                .flatMap(a -> Arrays.stream(a))
                .map(s -> Integer.parseInt(s))
                .filter(integer -> integer % 2 == 0 && integer != 1)
                .min(Comparator.naturalOrder())
                .orElseThrow(() -> new RuntimeException("Can't get min value from list:" + numbers));
```

olekskov (changes requested):
Good job overall, please fix codestyle according to comments

olekskov (approved):
Good job

PR #951, opened 2023-11-23
Task 1:

Alex-Solod:
Extract minimum even integer from a list of strings containing multiple numbers

Task 2:
Subtract 1 from odd-indexed elements and calculate average of odd numbers

Task 3:
Select men of army recruitment age (18-27 years) from a list of People

Task 4:
Select people of working age (18-60 for men, 18-55 for women) from a list of People

Task 5:
Return names of cats owned by women of a certain age from a list of People

Task 6:
Validate candidates for president: age ≥ 35, allowed to vote, Ukrainian nationality, and lived in Ukraine for 10 years Also implemented a Predicate parametrized CandidateValidator method

### DIFF
--- a/src/main/java/practice/CandidateValidator.java
+++ b/src/main/java/practice/CandidateValidator.java
@@ -1,5 +1,22 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    @Override
+    public boolean test(Candidate candidate) {
+        if (candidate.getAge() >= 35
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals("Ukrainian")
+                && candidate.getPeriodsInUkr() != null) {
+            String[] periods = candidate.getPeriodsInUkr().split("-");
+            if (periods.length == 2) {
+                int startYear = Integer.parseInt(periods[0]);
+                int endYear = Integer.parseInt(periods[1]);
+                return endYear - startYear >= 10;
+            }
+        }
+        return false;
+    }
 }
--- a/src/main/java/practice/StreamPractice.java
+++ b/src/main/java/practice/StreamPractice.java
@@ -1,8 +1,13 @@
 package practice;
 
-import java.util.Collections;
+import java.util.Arrays;
+import java.util.Comparator;
 import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import model.Candidate;
+import model.Cat;
 import model.Person;
 
 public class StreamPractice {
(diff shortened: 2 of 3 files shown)


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,22 @@
...
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    @Override
+    public boolean test(Candidate candidate) {
+        if (candidate.getAge() >= 35
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals("Ukrainian")
```
kozhukhovsky, line 11: move string literals and indexes to the constant

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(s -> Arrays.stream(s.split(",")))
+                .filter(n -> Integer.parseInt(n) % 2 == 0)
```
kozhukhovsky, line 24:
```suggestion
                .filter(number -> Integer.parseInt(n) % 2 == 0)
```

I would recommend not using an abbreviation if you can give the full name

kozhukhovsky (changes requested):

src/main/java/practice/CandidateValidator.java
Alex-Solod: Got u

src/main/java/practice/StreamPractice.java
Alex-Solod: ok ok

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +69,14 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
...
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> (p.getSex().equals(Person.Sex.WOMAN)
+                        && p.getAge() >= fromAge
+                        && p.getAge() <= femaleToAge)
+                        || (p.getSex().equals(Person.Sex.MAN)
+                                && p.getAge() >= fromAge
+                                && p.getAge() <= maleToAge))
```
olekskov, line 78: like here

olekskov (approved):
Approved but further please avoid 1-2 letter abbreviations for variables naming even when using streams.
It makes code much harder to read when you forget context of what is going on

src/main/java/practice/StreamPractice.java
Alex-Solod: Got it thank u)

PR #953, opened 2023-11-24
stream_practice_firstCommit

OlegShvets1:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,20 @@
...
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MIN_AGE = 35;
+    private static final int MIN_RESIDENCY_YEARS = 10;
+    private static final String REQUIRED_NATIONALITY = "Ukrainian";
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] yearsRange = candidate.getPeriodsInUkr().split("-");
+        int stayDuration = Integer.parseInt(yearsRange[1]) - Integer.parseInt(yearsRange[0]);
+        return candidate.isAllowedToVote()
```
okuzan, line 15: 0 1 "-" should be constants

src/main/java/practice/StreamPractice.java
```diff
@@ -23,9 +33,16 @@ public int findMinEvenNumber(List<String> numbers) {
...
+        return IntStream.range(0, numbers.size())
+                .map(index -> !isNumberEven(index) ? numbers.get(index) - 1 : numbers.get(index))
+                .filter(i -> !isNumberEven(i))
+                .average()
+                .orElseThrow(NoSuchElementException::new);
     }
 
+    private boolean isNumberEven(int number) {
+        return number % 2 == 0;
+    }
```
okuzan, line 45: place private method after public ones

okuzan (changes requested):

olekskov (approved):
Good job

PR #963, opened 2023-12-05
Hw stream

AGuliaiev:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,78 @@
...
+        return numbers.stream()
+                .flatMapToInt(s -> {
+                    String[] split = s.split(",");
+                    int[] ints = new int[split.length];
+                    for (int i = 0; i < split.length; i++) {
+                        ints[i] = Integer.parseInt(split[i].trim());
+                    }
+                    return java.util.Arrays.stream(ints);
+                })
+                .filter(i -> i % 2 == 0)
```
kozhukhovsky, line 26:
```suggestion
        return numbers.stream()
                .flatMap(string -> Stream.of(string.split(","))).
                .mapToInt(Integer::parseInt)
                .filter(i -> i % 2 == 0)
```

kozhukhovsky: and `i `bad practice for name, rename pls

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,78 @@
...
-     * parametrized with Candidate in CandidateValidator.
-     */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(new CandidateValidator())
+                .filter(candidate ->
+                        Integer.parseInt(candidate.getPeriodsInUkr().split("-")[1])
+                                - Integer.parseInt(candidate.getPeriodsInUkr()
+                                .split("-")[0]) >= PERIODS_IN_UKR)
```
kozhukhovsky, line 70: extract here to the CandidateValidator

src/main/java/practice/CandidateValidator.java
```diff
@@ -6,12 +6,15 @@
...
 
     @Override
     public boolean test(Candidate candidate) {
         return candidate.getAge() >= MIN_AGE
                 && candidate.isAllowedToVote()
                 && NATIONALITY.equals(candidate.getNationality())
-                && candidate.getPeriodsInUkr().matches("\\d{4}-\\d{4}");
+                && candidate.getPeriodsInUkr().matches("\\d{4}-\\d{4}")
+                && (Integer.parseInt(candidate.getPeriodsInUkr().split("-")[1])
+                - Integer.parseInt(candidate.getPeriodsInUkr().split("-")[0]) >= PERIODS_IN_UKR);
```
kozhukhovsky, line 18: better move to the constant splitter and indexes

kozhukhovsky (approved):

PR #967, opened 2023-12-10
added solution

furthernull:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(s -> Arrays.stream(s.split(",")))
+                .map(Integer::parseInt)
```
Sanderfuck, line 23: You can use mapToInt method

Sanderfuck (approved):

PR #970, opened 2023-12-12
resolved tasks using Stream API

YaroslavRadevychVynnytskyi:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,13 @@ public class StreamPractice {
...
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        OptionalInt min = numbers.stream()
+                .flatMap(str -> Arrays.stream(str.split(",")))
+                .mapToInt(Integer::parseInt)
+                .filter(number -> number % 2 == 0)
+                .min();
+        return min.orElseThrow(() -> new RuntimeException("Can't get min value from list:"
+                + " < Here is our input " + numbers + " >"));
```
kerrrusha, line 29:
```suggestion
        return numbers.stream()
                .flatMap(str -> Arrays.stream(str.split(",")))
                .mapToInt(Integer::parseInt)
                .filter(number -> number % 2 == 0)
                .min()
                .orElseThrow(() -> new RuntimeException("Can't get min value from list:"
                + " < Here is our input " + numbers + " >"));
```

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +35,11 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        OptionalDouble average = IntStream.range(0, numbers.size())
+                .map(i -> i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i))
+                .filter(number -> number % 2 != 0)
+                .average();
+        return average.orElseThrow(() -> new NoSuchElementException("There is no such element!"));
```
kerrrusha, line 42:
```suggestion
        return IntStream.range(0, numbers.size())
                .map(i -> i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i))
                .filter(number -> number % 2 != 0)
                .average()
                .orElseThrow(() -> new NoSuchElementException("There is no such element!"));
```

kerrrusha (approved):
good job but can be improved a little

PR #974, opened 2023-12-18
created methods

KhrystynaDidykh:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,15 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        Optional<Integer> number = numbers
+                .stream()
+                .flatMap(s -> Arrays.stream(s.split(",")))
```
kozhukhovsky, line 25: It’s better not to give one-letter names, even on streams

kozhukhovsky (approved):

PR #975, opened 2023-12-19
Realized methods using Stream API

SSosedova:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,18 @@
...
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String periodsInUkr = candidate.getPeriodsInUkr();
+        String[] periods = periodsInUkr.split("-");
+        int yearsInUkr = Integer.parseInt(periods[1]) - Integer.parseInt(periods[0]);
+        return candidate.getAge() >= 35
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals("Ukrainian")
+                && yearsInUkr >= 10;
```
okuzan, line 16: extract all literals to constants

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,18 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+
```
```suggestion
```

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +17,15 @@ public class StreamPractice {
...
-        return 0;
+        return numbers.stream()
+                .map(s -> s.split(","))
+                .flatMap(Stream::of)
+                .map(Integer::valueOf)
+                .filter(num -> num % 2 == 0)
+                .sorted()
+                .findFirst()
+                .orElseThrow(() ->
+                        new RuntimeException("Can't get min value from list: " + numbers));
```
okuzan, line 28: throw NoSuchElement in such cases

okuzan (changes requested):
don't use short one-letter names for variables even in stream chains (outside of math context), it decreases code readability, extract string, integer literals to constants

kozhukhovsky (approved):

PR #977, opened 2023-12-25
Implemented StreamPractice and CandidateValidator classes.

davydov-software:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,15 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        Supplier<RuntimeException> exceptionSupplier = ()
+                -> new RuntimeException("Can't get min value from list: " + numbers);
+        return numbers.stream()
+                .map(s -> Arrays.asList(s.split(",")))
+                .flatMap(Collection::stream)
```
olekskov, line 27: .map + .flatMap right after it usually can be replaced with just .flatMap
```suggestion
                .flatMap(s -> Arrays.stream(s.split(",")))
```

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +37,11 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream
+                .range(0, numbers.size())
+                .map(i -> numbers.get(i) - i % 2)
+                .filter(i -> i % 2 != 0)
+                .average().getAsDouble();
```
olekskov, line 44: Do not combine operations in one row when others are single. It's not expected and can be missed during code reading
```suggestion
                .average()
                .getAsDouble();
```

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +89,13 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex() == Person.Sex.WOMAN && p.getAge() >= femaleAge)
+                .map(Person::getCats)
+                .flatMap(Collection::stream)
```
olekskov, line 95: Please replace 2 commands (.map+.flatMap) with just proper .flatMap()

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +53,13 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        Predicate<Person> predicate = p
+                -> p.getSex() == Person.Sex.MAN
+                && p.getAge() >= fromAge
+                && p.getAge() <= toAge;
```
olekskov, line 59: Please replace all 1-letter abbreviations in code for proper readable names.
You do not need to save space here but after some time (when you forget what is p, i or any other letter is) code becomes unreadable

olekskov (changes requested):

olekskov (approved):
Good job

PR #981, opened 2024-01-06
done

163onmyneck:


kerrrusha (changes requested):

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMapToInt(s -> Arrays.stream(s.split(","))
+                        .mapToInt(Integer::valueOf))
```
Sanderfuck, line 25: You just can use flatMap method and after this mapToInt, it looks better

Sanderfuck (approved):

PR #991, opened 2024-01-18
task done)

AndrewOliferchuk:


okuzan (approved):
Avoid one letter naming in stream chains

PR #992, opened 2024-01-20
stream practice completed

bogdankost:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,15 @@
...
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] periodInUkr = candidate.getPeriodsInUkr().split("-");
+        return candidate.getAge() >= 35
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals("Ukrainian")
+                && (Integer.parseInt(periodInUkr[1]) - Integer.parseInt(periodInUkr[0]) >= 10);
```
okuzan, line 13: please make constants from integer/string literals

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +50,11 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex() == Person.Sex.MAN
```
okuzan, line 54: p -> person, don't neglect informative naming even in stream chains

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,14 @@ public class StreamPractice {
...
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(n -> n.split(","))
+                .flatMap(Arrays::stream)
+                .mapToInt(Integer::parseInt)
+                .filter(n -> n % 2 == 0)
+                .min()
+                .orElseThrow(()
+                        -> new RuntimeException("Can't get min value from list:" + numbers));
```
okuzan, line 28: NoSuchElement

okuzan (changes requested):

src/main/java/practice/StreamPractice.java
bogdankost: > NoSuchElement

but the condition said 'If there is no needed data throw RuntimeException with message "Can't get min value from list: < Here is our input 'numbers' >"'

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,21 @@
...
+    private static final int MIN_PERIOD = 10;
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] periodInUkr = candidate.getPeriodsInUkr().split(SPLITERATOR);
+        return candidate.getAge() >= MIN_AGE
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals(REQUIRED_NATIONALITY)
+                && (Integer.parseInt(periodInUkr[1])
+                - Integer.parseInt(periodInUkr[0]) >= MIN_PERIOD);
```
okuzan, line 19: PERIOD_START_INDEX +  PERIOD_END_INDEX

src/main/java/practice/StreamPractice.java
okuzan: Sorry! my bad

okuzan (approved):

PR #997, opened 2024-01-29
dsumaroka/jv-stream-practice

dsumaroka:


src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +39,21 @@ public int findMinEvenNumber(List<String> numbers) {
...
+
+        Stream<Integer> streamEven = IntStream.range(0, numbers.size())
+                .filter(i -> i % 2 == 0)
+                .mapToObj(numbers::get);
+
+        return Stream.concat(streamEven, streamOdd)
+                .filter(n -> n % 2 != 0)
+                .mapToDouble(n -> n)
+                .average()
+                .orElseThrow();
```
olekskov, line 55: Please try to make in one stream. You're correct that we need to use range.
But next we can use .mapToDouble method with ternary operator inside to get element from list or element - 1 to combine a lot of actions

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,17 @@ public class StreamPractice {
...
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        try {
+            return numbers.stream()
+                    .map(s -> s.split(","))
+                    .flatMap(Arrays::stream)
+                    .mapToInt(Integer::valueOf)
+                    .filter(n -> n % 2 == 0)
+                    .reduce(Integer::min)
+                    .getAsInt();
```
olekskov, line 30: We shouldn't use 1-2 letter abbreviations for variables even in stream. Lines are already short, normal full naming will only make it more readable

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,17 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        try {
+            return numbers.stream()
+                    .map(s -> s.split(","))
+                    .flatMap(Arrays::stream)
```
olekskov, line 26: map+flatmap usually can be replaced with just flatmap

src/main/java/practice/StreamPractice.java
olekskov, line 30:
```suggestion
                    .min()
                    .orElseThrow(...some proper exception here..);
```

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,17 @@ public class StreamPractice {
...
+                    .mapToInt(Integer::valueOf)
+                    .filter(n -> n % 2 == 0)
+                    .reduce(Integer::min)
+                    .getAsInt();
+        } catch (NoSuchElementException e) {
+            throw new NoSuchElementException("Can't get min value from list");
```
olekskov, line 32: and then try/catch is not required at all

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +96,12 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex().equals(Person.Sex.WOMAN) && p.getAge() >= femaleAge)
+                .map(Person::getCats)
+                .flatMap(Collection::stream)
```
olekskov, line 102: map+flatmap -> one proper flatmap

olekskov (changes requested):

dsumaroka:
Ок

Пн, 29 янв. 2024 г. в 21:26, Oleksandr Kovalenko ***@***.***>:

PR #999, opened 2024-01-30
My stream implement

petro2061:
In this task implemented all methods form class StreamPractice use stream API. Realized custom implementation for CandidateValidator class use for it function interface Predicate.


kozhukhovsky (approved):
Good job!

PR #1000, opened 2024-01-31
solvedTask

shumYevh:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,20 @@
...
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= 35
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals("Ukrainian")
+                && getIntPeriodInUkr(candidate.getPeriodsInUkr()) >= 10;
+    }
+
+    //Only for String type like: "2002-2015"
+    public int getIntPeriodInUkr(String periodInUkr) {
```
okuzan, line 16: private

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,20 @@
...
+        String[] period = periodInUkr.split("-");
+        return Integer.parseInt(period[1]) - Integer.parseInt(period[0]);
```
okuzan, line 18: extract all literals, integers, strings, to constants

okuzan (changes requested):

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,75 @@
...
-        return 0D;
+        List<Integer> modifiedList = IntStream.range(0,numbers.size())
+                .mapToObj(i -> i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i))
+                .toList();
+
+        return modifiedList.stream()
+                .filter(el -> el % 2 != 0)
+                .mapToInt(Integer::intValue)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
olekskov, line 36: 1) Why we collect stream to open it again?
2) Why mapToObj when we may proceed to work with numbers?
```suggestion
        return IntStream.range(0,numbers.size())
                .map(i -> i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i))
                .filter(el -> el % 2 != 0)
                .average()
                .orElseThrow(NoSuchElementException::new);
```

olekskov (approved):

PR #1003, opened 2024-02-07
added solution to the task

yanna-stepanova:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +28,14 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(s -> s.split(COMMA_REGEX))
+                .flatMap(Arrays::stream)
+                .flatMapToInt(n -> IntStream.of(Integer.parseInt(n)))
+                .filter(i -> i % NUM_CHECK_ODD == NUM_REMAINDER)
+                .min()
```
olekskov, line 36: 1) .map + .flatmap are usually can be replaced with just one proper .flatmap() method
2) Please do no use 1-2 letter abbreviations for variables, also in streams. Lines are already short, you gain nothing using abbreviation but makes code less readable when you don't remember what is n, s, o, i, p, or any other letter

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +44,29 @@ public int findMinEvenNumber(List<String> numbers) {
...
+                                    if (Integer.parseInt(
+                                            String.valueOf(
+                                                    e.charAt(INDEX_FIRST))) % NUM_CHECK_ODD
+                                                                            != NUM_REMAINDER) {
+                                        value--;
+                                    }
+                                    return value;
+                                }));
+        return mapsOfInt.values()
+                .stream()
```
olekskov, line 64: to simplify solution use range and mapToDouble operation instead of this whole structure
IntStream.range(0,numbers.size())

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +44,29 @@ public int findMinEvenNumber(List<String> numbers) {
...
+                .filter(e -> e % NUM_CHECK_ODD != NUM_REMAINDER).toList()
+                .stream()
```
olekskov, line 66: Why collect and then again to stream? :)
```suggestion
                .filter(e -> e % NUM_CHECK_ODD != NUM_REMAINDER)
```

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +44,29 @@ public int findMinEvenNumber(List<String> numbers) {
...
+                .mapToDouble(Integer::doubleValue)
```
olekskov, line 67: after changes it shouldn't be required
```suggestion
```

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +109,13 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getSex() == Person.Sex.WOMAN
+                        && person.getAge() >= femaleAge)
+                .map(Person::getCats)
+                .flatMap(Collection::stream)
```
olekskov, line 116: map + flatmap = just flatmap
```suggestion
                .flatmap(person -> person.getCats().stream())
```

olekskov (changes requested):

okuzan (approved):
Good job!

PR #1005, opened 2024-02-08
hw-stream

LizaBalshv:


src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +34,12 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(index -> index % 2 != 0 ? numbers.get(index) - 1 : numbers.get(index))
+                .filter(num -> num % 2 != 0)
+                .mapToDouble(Double::valueOf)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
olekskov, line 42: if we have some mapping before we often may combine it with type change
```suggestion
        return IntStream.range(0, numbers.size())
                .mapToDouble(index -> index % 2 != 0 ? numbers.get(index) - 1 : numbers.get(index))
                .filter(num -> num % 2 != 0)
                .average()
                .orElseThrow(NoSuchElementException::new);
```

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(s -> Arrays.stream(s.split(",")))
+                .map(Integer::parseInt)
+                .filter(n -> n % 2 == 0)
+                .min(Integer::compareTo)
```
olekskov, line 26: similar approach here
```suggestion
                .mapToInt(Integer::parseInt)
                .filter(n -> n % 2 == 0)
                .min()
```

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +51,10 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(a -> a.getAge() > fromAge
+                        && a.getAge() <= toAge && a.getSex() == Person.Sex.MAN)
```
olekskov, line 56: Please do not use 1-2 letter abbreviations for variable names. Even in streams.
In a week nobody will remember what is a, s, n, i or any other letter
Please recheck all methods

olekskov (changes requested):
Good job, just some style comments

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(s -> Arrays.stream(s.split(",")))
+                .mapToInt(Integer::parseInt)
+                .filter(n -> n % 2 == 0)
```
okuzan, line 25: n -> num

okuzan (approved):
well done

PR #1007, opened 2024-02-12
🤔If only i could solve all tasks overnight🤔

orest40th:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,22 @@
...
+    public boolean test(Candidate candidate) {
+        int residenceYears = Arrays.stream(candidate.getPeriodsInUkr()
+                        .split(CANDIDATE_VALIDATOR_EX))
+                .mapToInt(Integer::parseInt)
+                .reduce((first, second) -> second - first)
+                .orElseThrow(() -> new IllegalArgumentException("Invalid range format"));
+
+        return candidate.getAge() >= 35 && candidate.isAllowedToVote()
+                && candidate.getNationality().equals("Ukrainian")
+                && residenceYears >= 10;
```
okuzan, line 20: image

okuzan (changes requested):

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,24 @@
...
+    @Override
+    public boolean test(Candidate candidate) {
+        int residenceYears = Arrays.stream(candidate.getPeriodsInUkr()
+                        .split(CANDIDATE_VALIDATOR_EX))
+                .mapToInt(Integer::parseInt)
+                .reduce((first, second) -> second - first)
+                .orElseThrow(() -> new IllegalArgumentException("Invalid range format"));
+
+        return candidate.getAge() >= MIN_AGE && candidate.isAllowedToVote()
+                && candidate.getNationality().equals("Ukrainian")
```
okuzan, line 21: string literal to constant

okuzan (approved):

PR #1018, opened 2024-03-02
implemnted all methods in Stream Practice and CandidateValidator classes

Oleksii-Naumenko:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,20 +1,28 @@
...
-     */
+    private static final int EVEN_OR_ODD_NUMBER_DIVIDER = 2;
+    private static final int DECREMENT_ONE = 1;
+    private static final String DELIMETER = ",";
+
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(string -> string.split(DELIMETER))
+                .flatMap(Arrays::stream)
```
olekskov, line 20: map + flatMap usually can be replaced with just proper flatMap

src/main/java/practice/StreamPractice.java
```diff
@@ -23,58 +31,52 @@ public int findMinEvenNumber(List<String> numbers) {
...
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(index -> {
+                    if (index % EVEN_OR_ODD_NUMBER_DIVIDER != 0) {
+                        return numbers.get(index) - DECREMENT_ONE;
+                    }
+                    return numbers.get(index);
+                })
+                .mapToDouble(Double::valueOf)
```
olekskov, line 41:
```suggestion
                .mapToDouble(index -> {
                    if (index % EVEN_OR_ODD_NUMBER_DIVIDER != 0) {
                        return numbers.get(index) - DECREMENT_ONE;
                    }
                    return numbers.get(index);
                })
```

olekskov (changes requested):

olekskov (approved):
Good job

PR #1019, opened 2024-03-02
added hw_solution

SkillfulDev:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,25 @@
...
+    public static final String DASH = "-";
+    public static final int LIVING_YEARS_IN_UKR = 10;
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] splitPeriod = candidate.getPeriodsInUkr().split(DASH);
+        Integer yearsLivedInUkraine = Arrays.stream(splitPeriod)
+                .map(Integer::valueOf)
+                .reduce((starYear, endYear) -> endYear - starYear)
+                .orElseThrow();
```
k0sm0naft, line 19: I'm not sure it's the best solution, but it's OK for practice.

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +21,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(element -> Arrays.stream(element.split(COMMA)))
+                .map(Integer::valueOf)
+                .filter(number -> number % DIVISOR_FOR_EVEN_CHECK == 0)
+                .min(Integer::compare)
```
k0sm0naft, line 28: I don't think it's wrong, but I think it's better.
```suggestion
                .mapToInt(Integer::valueOf)
                .filter(number -> number % DIVISOR_FOR_EVEN_CHECK == 0)
                .min()
```

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +73,12 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> (person.getSex() == Person.Sex.MAN
+                        && person.getAge() >= fromAge && person.getAge() <= maleToAge)
+                        || (person.getSex() == Person.Sex.WOMAN
+                        && person.getAge() >= fromAge && person.getAge() <= femaleToAge))
```
k0sm0naft, line 80: Group the common conditions.

k0sm0naft (changes requested):
You only need to correct the comment about conditions, the rest is up to you.

src/main/java/practice/StreamPractice.java
SkillfulDev: changed

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,25 @@
...
-    //write your code here
+import java.util.Arrays;
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    public static final int REQUIRED_AGE = 35;
+    public static final String NATIONALITY = "Ukrainian";
+    public static final String DASH = "-";
+    public static final int LIVING_YEARS_IN_UKR = 10;
```
andrii-hoienko, line 11:
```suggestion
    private static final int REQUIRED_AGE = 35;
    private static final String NATIONALITY = "Ukrainian";
    private static final String DASH = "-";
    private static final int LIVING_YEARS_IN_UKR = 10;
```

src/main/java/practice/StreamPractice.java
```diff
@@ -1,11 +1,18 @@
...
 import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import model.Candidate;
+import model.Cat;
 import model.Person;
 
 public class StreamPractice {
+    public static final String COMMA = ",";
```
andrii-hoienko, line 13:
```suggestion
    private static final String COMMA = ",";
```

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +21,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(element -> Arrays.stream(element.split(COMMA)))
```
andrii-hoienko, line 25: Is it possible to 'flatMapToInt'?

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +36,14 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        int initialNumber = 0;
+        int subtractNumber = 1;
+        return IntStream.range(initialNumber, numbers.size())
+                .map(index -> index % DIVISOR_FOR_EVEN_CHECK == 0 ? numbers.get(index)
```
andrii-hoienko, line 42: It would be great if you will subtract content of map to separate method

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +36,14 @@ public int findMinEvenNumber(List<String> numbers) {
...
+                        : numbers.get(index) - subtractNumber)
+                .filter(element -> element % DIVISOR_FOR_EVEN_CHECK != 0)
```
andrii-hoienko, line 44: Code duplication

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +86,12 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getSex() == Person.Sex.WOMAN
```
andrii-hoienko, line 90: Check for code duplication. If methods in stream operations are long you can move them to separate methods

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +107,10 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(new CandidateValidator())
```
andrii-hoienko, line 111: Is it expected to create CandidateValidator each time you call this method?

andrii-hoienko (changes requested):
Good job, but some improvements required

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +24,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMapToInt(element -> Arrays.stream(element.split(COMMA))
```
andrii-hoienko, line 28: I believe after 'flatMapToInt' no reason to cast it again by 'mapToInt'

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +105,38 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
...
+        return candidates.stream()
+                .filter(candidateValidator)
+                .map(Candidate::getName)
+                .sorted()
+                .collect(Collectors.toList());
+    }
+
+    private IntUnaryOperator conditionalSubtractOrOriginal(List<Integer> numbers,
+                                                           int subtractNumber) {
+        return index -> index % DIVISOR_FOR_EVEN_CHECK == 0 ? numbers.get(index)
```
andrii-hoienko, line 118: index % DIVISOR_FOR_EVEN_CHECK == 0 this code duplication is not removed

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +105,38 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
...
+    }
+
+    private Predicate<Person> getPersonPredicate(int fromAge,
+                                                       int femaleToAge, int maleToAge) {
+        return person -> person.getAge() >= fromAge
+                && ((person.getSex() == Person.Sex.MAN && person.getAge() <= maleToAge)
+                || (person.getSex() == Person.Sex.WOMAN && person.getAge() <= femaleToAge));
+    }
+
+    private Predicate<Person> getPersonPredicateForWoman(int femaleAge) {
```
andrii-hoienko, line 129: I believe this method should return boolean and be named like 'isValidWoman' (this is bad name think about better one) it is not common practice to return predicates or other functional interfaces

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +105,38 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        CandidateValidator candidateValidator = new CandidateValidator();
```
andrii-hoienko, line 108: Should be class level field

andrii-hoienko (changes requested):
Nice, but some more improvements required

PR #1024, opened 2024-03-07
main version

EduarfF:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,19 @@ public class StreamPractice {
...
+                .filter(number -> number % 2 == 0)
+                .toList();
+
+        if (evenNumbers.isEmpty()) {
+            throw new RuntimeException("Can't get min value from list" + numbers);
+        }
+
+        return evenNumbers.stream()
+                .min(Comparator.naturalOrder())
+                .orElseThrow(NoSuchElementException::new);
```
karma-o, line 34: we need to have IntStream to be able to use it's methods without a comparator and not split the stream into 2 parts 
```suggestion
                .mapToInt(Integer::parseInt)
                .filter(number -> number % 2 == 0)
                .min()
                .orElseThrow(() -> new RuntimeException(("Can't get min value from list" + numbers));
```

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +40,15 @@ public int findMinEvenNumber(List<String> numbers) {
...
-        return 0D;
+        List<Integer> oddNumbers = IntStream.range(0, numbers.size())
+                .mapToObj(i -> i % 2 == 1 ? numbers.get(i) - 1 : numbers.get(i))
+                .toList();
+
+        return oddNumbers.stream()
+                .filter(i -> i % 2 == 1)
+                .mapToInt(Integer::intValue)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
karma-o, line 51: no need to split it into 2 streams let's try to keep it in a single one

karma-o (approved):
Approved with some minor comments, please take a look

PR #1025, opened 2024-03-08
implemented all methodsÐ

Bohadan-Markatov:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,25 @@
...
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] dividedDate = candidate.getPeriodsInUkr().split(YEAR_RANGE_SEPARATOR);
+        int dateFrom = Integer.parseInt(dividedDate[YEAR_FROM_COLUMN]);
+        int dateTo = Integer.parseInt(dividedDate[YEAR_TO_COLUMN]);
+
+        return (dateTo - dateFrom >= MIN_ACCEPTABLE_NUMBERS_OF_YEARS)
+                && (candidate.getAge() >= MIN_ACCEPTABLE_AGE)
+                && (candidate.getNationality().equals(ACCEPTABLE_NATIONALITY))
```
oksana-divkit, line 22: What if `candidate.getNationality()` returns `null`?

src/main/java/practice/StreamPractice.java
```diff
@@ -14,28 +21,48 @@ public class StreamPractice {
...
     /**
      * Given a List of Integer numbers,
      * return the average of all odd numbers from the list or throw NoSuchElementException.
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
+
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        OptionalDouble optionalDouble
+                = IntStream.iterate(0, i -> i < numbers.size(), i -> i + 1)
```
oksana-divkit, line 43: There is another method `IntStream.range(0, 10)`, could be simplier

src/main/java/practice/StreamPractice.java
```diff
@@ -14,28 +21,48 @@ public class StreamPractice {
...
+                .map(i -> i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i))
+                .filter(i -> i % 2 != 0)
+                .average();
+        return optionalDouble.orElseThrow(NoSuchElementException::new);
```
oksana-divkit, line 47: Here we can use just `getAsDouble()`. It itself throws `NoSuchElementException`.
```suggestion
                .average();
        return optionalDouble.getAsDouble();
```

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +116,10 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(c -> new CandidateValidator().test(c))
```
oksana-divkit, line 120: It is possible not to call the method 'test' itself, and simply pass the predicate
```suggestion
                .filter(new CandidateValidator())
```

oksana-divkit (approved):
As always, great work!

src/main/java/practice/CandidateValidator.java
Bohadan-Markatov: I think this is the responsibility of the Candidate class, but I agree that there are no checks and the value null can come

PR #1030, opened 2024-03-12
Implemented using streams

TonyH277:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,20 +1,34 @@
...
+                    .map(s -> s.split(","))
+                    .flatMap(s -> Arrays.stream(s))
+                    .map(Integer::valueOf)
+                    .filter(n -> n % 2 == 0)
+                    .mapToInt(Integer::valueOf)
+                    .min()
+                    .getAsInt();
+        } catch (RuntimeException e) {
+            throw new RuntimeException("Can't get min value from list: method_input_list " + e);
+        }
```
VlasenkoYevhenii, line 31: The try-catch construction is not needed here - just use .orElseThrow()
```suggestion
        return numbers.stream()
                .map(s -> s.split(","))
                .flatMap(Arrays::stream)
                .map(Integer::parseInt)
                .filter(n -> n % 2 == 0)
                .sorted()
                .findFirst()
                .orElseThrow(() -> new RuntimeException("Can't get min value from list: "
                                                                                  + numbers));
```

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +37,18 @@ public int findMinEvenNumber(List<String> numbers) {
...
+                numbers.set(i, numbers.get(i) - 1);
+            }
+            return numbers.stream()
+                    .filter(n -> n % 2 != 0)
+                    .mapToInt(Integer::valueOf)
+                    .average()
+                    .getAsDouble();
+        } catch (NoSuchElementException e) {
+            throw new NoSuchElementException("Can't find odd number" + e);
+        }
```
VlasenkoYevhenii, line 51: Same about try-catch here
Also, you're changing your input List - that's a bad idea
Don't use the for() cycle - it's not needed here, use IntStream.range(0, numbers.size()) instead (just google a little about stream iteration)
We don't need the message, just throw an exception by the task (but that's IMHO)

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +78,12 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> (p.getSex().equals(Person.Sex.MAN)
+                        && p.getAge() >= fromAge && p.getAge() <= maleToAge)
+                        || (p.getSex().equals(Person.Sex.WOMAN)
+                        && p.getAge() >= fromAge && p.getAge() <= femaleToAge))
```
VlasenkoYevhenii, line 85: I would split it into two checks
```suggestion
                .filter(p -> p.getAge() >= fromAge)
                .filter(p -> p.getSex().equals(Person.Sex.MAN) && p.getAge() <= maleToAge
                        || p.getSex().equals(Person.Sex.WOMAN) && p.getAge() <= femaleToAge)
```

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,16 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    @Override
```
VlasenkoYevhenii, line 7:
```suggestion
    private static final String VALID_NATIONALITY = "Ukrainian";
    private static final String PERIOD_SPLIT_REGEX = "-";```

**Review by VlasenkoYevhenii — CHANGES_REQUESTED** [review]
Good job!
There are some comments to check.

File: src/main/java/practice/CandidateValidator.java [hunk 5]
```diff
@@ -1,5 +1,21 @@
...
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int AGE_BARRIER = 35;
+    private static final int AGE_LIVED = 10;
+    private static final String NATIONALITY = "Ukrainian";
+    private static final String YEAR_REGEX = "-";
+
+    @Override
+    public boolean test(Candidate c) {
+        int fromYear = Integer.parseInt(c.getPeriodsInUkr().split(YEAR_REGEX)[0]);
```
andrii-hoienko, line 14: Magic number

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,21 @@
...
+        int toYear = Integer.parseInt(c.getPeriodsInUkr().split(YEAR_REGEX)[1]);
```
andrii-hoienko, line 15: Magic number

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,21 @@
...
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int AGE_BARRIER = 35;
+    private static final int AGE_LIVED = 10;
+    private static final String NATIONALITY = "Ukrainian";
+    private static final String YEAR_REGEX = "-";
+
+    @Override
+    public boolean test(Candidate c) {
```
andrii-hoienko, line 13:
```suggestion
    public boolean test(Candidate candidate ) { //Don`t use one word variables
```

src/main/java/practice/StreamPractice.java
```diff
@@ -1,20 +1,32 @@
...
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
+
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(s -> s.split(","))
+                .flatMap(s -> Arrays.stream(s))
+                .map(Integer::valueOf)
+                .filter(n -> n % 2 == 0)
```
andrii-hoienko, line 25: Code duplication

src/main/java/practice/StreamPractice.java
```diff
@@ -1,20 +1,32 @@
...
      * input = {"5,30,100", "0,22,7", ...}
      * return min integer value. One more thing - we're interested in even numbers.
      * If there is no needed data throw RuntimeException with message
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
+
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(s -> s.split(","))
```
andrii-hoienko, line 22: Move string literal to constant

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +90,11 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex().equals(Person.Sex.WOMAN) && p.getAge() >= femaleAge)
```
andrii-hoienko, line 94: It would be great if you will subtract content of filter operation (or other operations if they are complex) to separate method (check other methods as well). It will significantly improve readability

andrii-hoienko (changes requested):

PR #1031, opened 2024-03-12
Implemented stream practice tasks

VolandevlodD:
- Added methods to find minimum even integer value, calculate average of odd numbers, and select specific individuals from lists based on age and gender criteria.
- Assisted with election validation by implementing CandidateValidator predicate for president position eligibility.
- Enhanced code readability and reusability.

### DIFF
--- a/src/main/java/practice/CandidateValidator.java
+++ b/src/main/java/practice/CandidateValidator.java
@@ -1,5 +1,27 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    public static final int START_YEAR_INDEX = 0;
+    public static final int END_YEAR_INDEX = 1;
+    public static final int RESIDENCY_TENURE_THRESHOLD = 10;
+
+    public static final int MINIMUM_ELIGIBLE_AGE = 35;
+
+    @Override
+    public boolean test(Candidate candidate) {
+        boolean isResidentForTenYears = false;
+        if (candidate.getPeriodsInUkr() != null) {
+            String[] years = candidate.getPeriodsInUkr().split("-");
+            isResidentForTenYears = (Integer.parseInt(years[END_YEAR_INDEX])
+                            - Integer.parseInt(years[START_YEAR_INDEX]))
+                    >= RESIDENCY_TENURE_THRESHOLD;
+        }
+        return candidate.getAge() >= MINIMUM_ELIGIBLE_AGE
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals("Ukrainian")
+                && isResidentForTenYears;
+    }
 }
--- a/src/main/java/practice/StreamPractice.java
+++ b/src/main/java/practice/StreamPractice.java
@@ -1,8 +1,14 @@
 package practice;
 
-import java.util.Collections;
+import java.util.Arrays;
+import java.util.Collection;
 import java.util.List;
+import java.util.Optional;
+import java.util.function.IntToDoubleFunction;
+import java.util.function.Predicate;
+import java.util.stream.IntStream;
 import model.Candidate;
+import model.Cat;
 import model.Person;
 
 public class StreamPractice {
@@ -13,8 +19,19 @@ public class StreamPractice {
      * If there is no needed data throw RuntimeException with message
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
-    public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+
+    public static int findMinEvenNumber(List<String> numbers) {
+        Optional<Integer> min = numbers.stream()
+                .map(s -> s.split(","))
+                .flatMap(Arrays::stream)
+                .map(Integer::parseInt)
+                .filter(num -> num % 2 == 0)
+                .min(Integer::compare);
+        if (min.isPresent()) {
+            return min.get();
+        } else {
+            throw new RuntimeException("Can't get min value from list: " + numbers);
+        }
     }
 
     /**
@@ -23,7 +40,13 @@ public class StreamPractice {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        IntToDoubleFunction intToDoubleMapper =
+                i -> i % 2 == 0 ? numbers.get(i) : numbers.get(i) - 1;
+        return IntStream.range(0, numbers.size())
+                .mapToDouble(intToDoubleMapper)
+                .filter(num -> num % 2 != 0)
+                .average()
+                .getAsDouble();
     }
 
     /**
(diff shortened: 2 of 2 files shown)


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,27 @@
...
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    public static final int START_YEAR_INDEX = 0;
+    public static final int END_YEAR_INDEX = 1;
+    public static final int RESIDENCY_TENURE_THRESHOLD = 10;
```
makaroshyna, line 9: Such a good naming!))

src/main/java/practice/StreamPractice.java
```diff
@@ -13,8 +19,19 @@ public class StreamPractice {
      * If there is no needed data throw RuntimeException with message
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
-    public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+
+    public static int findMinEvenNumber(List<String> numbers) {
+        Optional<Integer> min = numbers.stream()
+                .map(s -> s.split(","))
```
makaroshyna, line 25: I'm not 100% sure, but I think you should do a constant of a comma.

src/main/java/practice/StreamPractice.java
```diff
@@ -13,8 +19,19 @@ public class StreamPractice {
...
+                .flatMap(Arrays::stream)
+                .map(Integer::parseInt)
+                .filter(num -> num % 2 == 0)
+                .min(Integer::compare);
+        if (min.isPresent()) {
+            return min.get();
+        } else {
+            throw new RuntimeException("Can't get min value from list: " + numbers);
+        }
```
makaroshyna, line 34: Better thing to do is to add .orElseThrow() in a stream, and there add your exception.

makaroshyna (changes requested):
Very good! But I would recomend you to change how you throw an exception.

src/main/java/practice/CandidateValidator.java
VolandevlodD: I asked ChatGPT for help, that's why)))

src/main/java/practice/StreamPractice.java
VolandevlodD: It's used only once, and imho it's clear from the context that we are splitting our string by comma delimiter. I believe we can leave it this way.

src/main/java/practice/StreamPractice.java
VolandevlodD: Thank you, resolved.

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,27 @@
...
+
+    public static final int MINIMUM_ELIGIBLE_AGE = 35;
```
andrii-hoienko, line 11:
```suggestion
    private static final int START_YEAR_INDEX = 0;
    private static final int END_YEAR_INDEX = 1;
    private static final int RESIDENCY_TENURE_THRESHOLD = 10;
    private static final int MINIMUM_ELIGIBLE_AGE = 35;
```

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,27 @@
...
+
+    @Override
+    public boolean test(Candidate candidate) {
+        boolean isResidentForTenYears = false;
+        if (candidate.getPeriodsInUkr() != null) {
+            String[] years = candidate.getPeriodsInUkr().split("-");
+            isResidentForTenYears = (Integer.parseInt(years[END_YEAR_INDEX])
+                            - Integer.parseInt(years[START_YEAR_INDEX]))
+                    >= RESIDENCY_TENURE_THRESHOLD;
+        }
```
andrii-hoienko, line 21: Move this logic to separate method, and also make string literals like '-' and 'Ukrainian' constants

src/main/java/practice/StreamPractice.java
andrii-hoienko: Agree, please make comma constant

src/main/java/practice/StreamPractice.java
```diff
@@ -13,8 +18,16 @@ public class StreamPractice {
...
      */
-    public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+
+    public static int findMinEvenNumber(List<String> numbers) {
+        return numbers.stream()
+                .map(s -> s.split(","))
+                .flatMap(Arrays::stream)
+                .map(Integer::parseInt)
+                .filter(num -> num % 2 == 0)
```
andrii-hoienko, line 27: code duplication

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +109,10 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(new CandidateValidator())
```
andrii-hoienko, line 113: make CandidateValidator class level variable, to not create it each time method called

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +73,13 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        Predicate<Person> properAgeBySex = person -> person.getSex() == Person.Sex.MAN
+                && person.getAge() >= fromAge && person.getAge() <= maleToAge
+                || person.getSex() == Person.Sex.WOMAN
+                && person.getAge() >= fromAge && person.getAge() <= femaleToAge;
```
andrii-hoienko, line 79: It would be great if you will subtract content of filter operation (or other operations if they are complex) to separate method (check other methods as well). It will significantly improve readability

andrii-hoienko (changes requested):
Good job, but some improvements required

src/main/java/practice/StreamPractice.java
VolandevlodD: Do you mean that there's duplication in the mapping process?

src/main/java/practice/StreamPractice.java
VolandevlodD: Should I encapsulate the logic within my Predicate's anonymous class?

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +75,29 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
...
+            }
+
+            private boolean isAgeAboveMinimumRequired(Person person) {
+                return person.getAge() >= fromAge;
+            }
+
+            private boolean isMan(Person person) {
+                return person.getSex() == Person.Sex.MAN;
+            }
+        };
```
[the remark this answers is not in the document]
VolandevlodD: Ok🤔, I think I get it now. It's better to create a separate method instead of a custom Predicate.

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,31 @@
...
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= MINIMUM_ELIGIBLE_AGE
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals(NATIONALITY_UKRAINIAN)
+                && isResidentForTenYears(candidate);
+    }
+
+    private boolean isResidentForTenYears(Candidate candidate) {
+        if (candidate.getPeriodsInUkr() != null) {
```
karma-o, line 23: reverse this if, `if null - then return`

karma-o (approved):
Looks very nice! Just left a comment on the return early pattern

PR #1032, opened 2024-03-13
Solution Stream practice

ghost:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,70 @@
...
-     * new Person(«Helen», 42, Sex.WOMAN))`,
-     * select from the List only men whose age is from `fromAge` to `toAge` inclusively.
-     * <p>
-     * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
-     */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex().equals(Person.Sex.MAN))
+                .filter(p -> p.getAge() >= fromAge && p.getAge() <= toAge)
```
TonyH277, line 41: could improve code readability 
```suggestion
                .filter(p -> p.getSex().equals(Person.Sex.MAN) 
                && p.getAge() >= fromAge && p.getAge() <= toAge)
```

TonyH277 (approved):
LGFM!

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,22 @@
...
+    public boolean test(Candidate candidate) {
+        return candidate.isAllowedToVote()
+                && candidate.getAge() >= 35
+                && candidate.getNationality().equals(VALID_NATIONALITY)
+                && calculatePeriodInUkraine(candidate) >= 10;
+    }
+
+    private int calculatePeriodInUkraine(Candidate candidate) {
+        String[] stringDates = candidate.getPeriodsInUkr().split(PERIOD_SPLIT_REGEX);
+        return Integer.parseInt(stringDates[1]) - Integer.parseInt(stringDates[0]);
```
andrii-hoienko, line 20: Magic numbers

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,22 @@
...
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final String VALID_NATIONALITY = "Ukrainian";
+    private static final String PERIOD_SPLIT_REGEX = "-";
+
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.isAllowedToVote()
+                && candidate.getAge() >= 35
```
andrii-hoienko, line 13: Magic numbers

src/main/java/practice/StreamPractice.java
andrii-hoienko: Agree

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,70 @@
...
-     * or to `femaleToAge` (for women) inclusively.
-     * <p>
-     * Example: select people of working age
-     * (from 18 y.o. and to 60 y.o. for men and to 55 y.o. for women inclusively).
-     */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getAge() >= fromAge)
```
andrii-hoienko, line 48: Is is better to use one filter, boolean expressions could be united

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,70 @@
...
-     * have nationality - 'Ukrainian'
-     * and live in Ukraine for 10 years. For the last requirement use field periodsInUkr,
-     * which has following view: "2002-2015"
-     * We want to reuse our validation in future, so let's write our own impl of Predicate
-     * parametrized with Candidate in CandidateValidator.
-     */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(c -> new CandidateValidator().test(c))
```
andrii-hoienko, line 65: No need to create CandidateValidator each time we call this method, make CandidateValidator class level variable

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,70 @@
...
 
-    /**
-     * Given a List of `Person` instances (having `name`, `age`, `sex` and `cats` fields,
-     * and each `Cat` having a `name` and `age`),
-     * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
-     */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex().equals(Person.Sex.WOMAN) && p.getAge() >= femaleAge)
```
andrii-hoienko, line 56: I suggest creating separate method with a name like 'isValidWoman' and define boolean expression there, such an approach increases code readability because method name adds additional context (check other methods as well)

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,70 @@
...
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(i -> {
+                    if (i % 2 != 0) {
+                        return numbers.get(i) - 1;
+                    }
+                    return numbers.get(i);
+                })
+                .filter(n -> n % 2 != 0)
+                .average().orElseThrow(NoSuchElementException::new);
```
andrii-hoienko, line 35:
```suggestion
                .average()
                .orElseThrow(NoSuchElementException::new); //Better to add come message
```

andrii-hoienko (changes requested):

andrii-hoienko (changes requested):
Please resolve all comments

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,70 @@
...
-     * "Can't get min value from list: < Here is our input 'numbers' >"
-     */
+
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(s -> s.split(","))
+                .flatMap(Arrays::stream)
+                .map(Integer::parseInt)
+                .filter(n -> n % 2 == 0)
```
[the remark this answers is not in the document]
VlasenkoYevhenii: image
But each line returns a different result type

src/main/java/practice/StreamPractice.java
VlasenkoYevhenii: I think this contradicts this comment
image

src/main/java/practice/StreamPractice.java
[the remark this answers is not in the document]
karma-o: from what I can see, we can combine the first `map` with the `flatMap` operation, and maybe we can extract `n % 2 == 0` to some `isEven(int num)` and reuse it in `getOddNumsAverage()`, but I think we can also add `isOdd(int num)` along with It
Also if you have a question it is better to ask it in the chat on the platform, we rarely check replies here

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,74 @@
...
-     * and each `Cat` having a `name` and `age`),
-     * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
-     */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex().equals(Person.Sex.WOMAN)
+                            && p.getAge() >= femaleAge)
+                .map(Person::getCats)
+                .flatMap(Collection::stream)
```
karma-o, line 62: could be a single operation

karma-o (approved):
Looks Ok to me, left some comments, we could improve readability by by extracting some of the filter functions to a separate methods, but it is not critical

PR #1036, opened 2024-03-13
first commit

MolchanovAlexander:


nklimovych (changes requested):
Looks pretty good to me 👍.
Left you a few suggestions. I hope they will be helpful 🔥

nklimovych:
commit names are fire 🔥

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,27 @@
...
+    private static final int END_TERM_INDEX = 1;
+    private static final String ALLOWED_NATIONALITY = "Ukrainian";
+    private static final String DIVIDER = "-";
+
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.isAllowedToVote()
+                && candidate.getAge() >= ALLOWED_AGE
+                && candidate.getNationality().equals(ALLOWED_NATIONALITY)
+                && getLivingTime(candidate) >= ALLOWED_YEARS_OF_LIVING;
```
karma-o, line 19: this sounds very scary 😄 
I'm joking of course, but we need to change the method name  and constant name, add more context to it to indicate that this is not just about being alive, but about living in Ukraine

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,90 @@
...
 public class StreamPractice {
-    /**
-     * Given list of strings where each element contains 1+ numbers:
-     * input = {"5,30,100", "0,22,7", ...}
-     * return min integer value. One more thing - we're interested in even numbers.
-     * If there is no needed data throw RuntimeException with message
-     * "Can't get min value from list: < Here is our input 'numbers' >"
-     */
+    public static final String NO_ELEMENT_MSG = "No odd numbers in the list or empty list.";
+    public static final String CANT_GET_VAL_MSG = "Can't get min value from list: ";
```
karma-o, line 14: let's not shorten variable names and use full words

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,90 @@
...
 
-    /**
-     * Given a List of Integer numbers,
-     * return the average of all odd numbers from the list or throw NoSuchElementException.
-     * But before that subtract 1 from each element on an odd position (having the odd index).
-     */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .mapToObj(i -> conditionForOddNumsAverage(numbers, i))
```
karma-o, line 29: this is not just the condition, because this method does not return boolean, it returns a number, and is called not in the `filter()`, so let's rename this `conditionForOddNumsAverage()` to something morefitting or just move it's logic back to here since it is only used here

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,90 @@
...
+
+    private boolean conditionForCatNames(int femaleAge, Person p) {
+        return p.getAge() >= femaleAge && p.getSex() == Person.Sex.WOMAN;
+    }
+
+    private int conditionForOddNumsAverage(List<Integer> numbers, int i) {
+        return i % 2 != 0 ? numbers.get(i) - 1 : numbers.get(i);
+    }
+
+    private boolean isManByAgeMatch(int fromAge, int toAge, Person p) {
```
karma-o, line 75:
```suggestion
    private boolean isManByAgeMatchInclusive(int fromAge, int toAge, Person p) {
```

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,90 @@
...
+        return p.getAge() <= toAge
+                && p.getAge() >= fromAge
+                && p.getSex() == Person.Sex.MAN;
+    }
+
+    private boolean conditionForWorkablePeople(int fromAge, int femaleToAge,
```
karma-o, line 81: again, not very good method namejust naming it after the method name does not bring up much of a context to it, what if we want to use this method somewhere else? I'd name it something like `personFitsAge()`

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,90 @@
...
+    public static final String CANT_GET_VAL_MSG = "Can't get min value from list: ";
+    private static final String DIVIDER = ",";
+
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(n -> n.split(DIVIDER))
+                .flatMap(Arrays::stream)
+                .mapToInt(Integer::valueOf)
+                .filter(f -> f % 2 == 0)
```
karma-o, line 22: why `f`?

karma-o (changes requested):
Just some style issues

PR #1038, opened 2024-03-13
Added implementation for all practice methods.

nklimovych:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,25 @@
...
+    public static final int MIN_YEARS_IN_UKRAINE = 10;
+    public static final String DIVIDER = "-";
+    public static final int FIRST_ELEMENT = 0;
+    public static final int SECOND_ELEMENT = 1;
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] periods = candidate.getPeriodsInUkr().split(DIVIDER);
+        int startYear = Integer.parseInt(periods[FIRST_ELEMENT]);
+        int endYear = Integer.parseInt(periods[SECOND_ELEMENT]);
```
vikkushnir, line 18: not sure, maybe only for visibility, make it in separate method

src/main/java/practice/StreamPractice.java
```diff
@@ -74,7 +97,34 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
...
+                || (person.getSex().equals(Person.Sex.WOMAN) && person.getAge() <= femaleToAge));
+    }
+
+    private static boolean isMaleWithinAgeRange(int fromAge, int toAge, Person p) {
+        return p.getSex().equals(Person.Sex.MAN) && p.getAge() >= fromAge && p.getAge() <= toAge;
+    }
+
+    private static boolean isFemaleAgedAbove(int femaleAge, Person person) {
+        return person.getSex().equals(Person.Sex.WOMAN) && person.getAge() >= femaleAge;
+    }
```
vikkushnir, line 121: Good choice to use checks in separate methods, I didn't think of that :) just only, maybe, better to move them after all public methods

vikkushnir (approved):
LGTM, few comments, but not necessary

PR #1039, opened 2024-03-13
Created stream practice

Recdt:


VolandevlodD (approved):
Looks nice to me

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,37 @@
...
+        return age >= MINIMAL_AGE;
+    }
+
+    private boolean isAcceptableNationality(String nationality) {
+        return nationality.equals(REQUIRED_NATIONALITY);
+    }
+
+    private int parseData(String data) {
+        String[] separatedData = data.split(DATA_SEPARATOR);
+        return Integer.parseInt(separatedData[1]) - Integer.parseInt(separatedData[0]);
```
karma-o, line 30: there is no context on what this method does and what is that data, let's add it by using 0 and 1 as constants and probably having a var for these array elements named, to have an even better understanding, we could also rename a method for that purpose

src/main/java/practice/StreamPractice.java
```diff
@@ -1,20 +1,37 @@
...
      * If there is no needed data throw RuntimeException with message
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
+
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(s -> Arrays.stream(s.split(COMMA_SEPARATOR)))
+                .map(Integer::parseInt)
+                .filter(e -> e % 2 == 0)
```
karma-o, line 30: let's put this `e % 2 == 0`t o the `isEven()` method and reuse in others as well

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +40,15 @@ public int findMinEvenNumber(List<String> numbers) {
...
-        return 0D;
+        IntStream.range(0, numbers.size())
+                .filter(this::isOdd)
+                .forEach(i -> numbers.set(i, numbers.get(i) - 1));
+
+        return numbers.stream()
+                .filter(this::isOdd)
+                .mapToDouble(Integer::doubleValue)
+                .average()
+                .orElseThrow(() -> new NoSuchElementException("List doesn't have odd numbers!"));
```
karma-o, line 51: we should have a single stream here

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +78,12 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        Predicate<Person> sexGetAgeValidator = (p) -> p.getAge() >= fromAge
+                && p.getAge() <= (p.getSex() == Person.Sex.WOMAN ? femaleToAge : maleToAge);
```
karma-o, line 82: better to not hardcode values like method paramethers in the predicate, I'd say that the best approach would be to add a private method instead of a predicate, it can return a predicate or a boolean

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +92,12 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex().equals(Person.Sex.WOMAN) && p.getAge() >= femaleAge)
+                .map(Person::getCats)
+                .flatMap(Collection::stream)
```
karma-o, line 98: can be a single `flatMap` operation

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +113,15 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(CANDIDATE_VALIDATOR)
+                .sorted(Comparator.comparing(Candidate::getName))
+                .map(Candidate::getName)
```
karma-o, line 119:
```suggestion
                .map(Candidate::getName)
                .sorted()
```

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +113,15 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
...
+                .collect(Collectors.toList());
+
```
karma-o, line 121: redundant empty line

src/main/java/practice/StreamPractice.java
Recdt: Doesn't work because get Cats it's a list, so I have a stream with lists. And i nedd to make it in a single stream.

src/main/java/practice/StreamPractice.java
```diff
@@ -1,20 +1,35 @@
...
+import java.util.NoSuchElementException;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import model.Candidate;
+import model.Cat;
 import model.Person;
 
 public class StreamPractice {
+    private static final String COMMA_SEPARATOR = ",";
+    private static final CandidateValidator CANDIDATE_VALIDATOR = new CandidateValidator();
```
andrii-hoienko, line 15:
```suggestion
    private final CandidateValidator сandidateValidator = new CandidateValidator();
```

andrii-hoienko (changes requested):
Very nice, left super minor comment

andrii-hoienko (approved):
Good job!

PR #1040, opened 2024-03-14
jv-stream-practice solution

vikkushnir:


src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +66,11 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                         .filter(p -> p.getSex() == Person.Sex.MAN
+                            ? p.getAge() >= fromAge && p.getAge() <= maleToAge
+                            : p.getAge() >= fromAge && p.getAge() <= femaleToAge)
```
azzilian, line 72: defenetly prvatei boolean method needed here

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +79,11 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                         .filter(p -> p.getSex() == Person.Sex.WOMAN && p.getAge() >= femaleAge)
+                         .flatMap(f -> f.getCats().stream())
```
azzilian, line 84: same
private method + full names for p and f

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +48,10 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex() == Person.Sex.MAN
```
azzilian, line 52: Btw its better to use full name instead of 1 letter p -> people

src/main/java/practice/StreamPractice.java
azzilian, line 72: p - persone or people - full word

azzilian (approved):
I think some minors improvement better be made mostly its about new private method and use full names instead of 1 letter but everything is ok

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +100,33 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
...
+    }
+
+    private static Predicate<Person> isWorkablePeople(int fromAge, int femaleToAge, int maleToAge) {
+        return person -> person.getSex() == Person.Sex.MAN
+                ? person.getAge() >= fromAge && person.getAge() <= maleToAge
+                : person.getAge() >= fromAge && person.getAge() <= femaleToAge;
+    }
+
+    private static Predicate<Person> isFemaleAgeAbove(int femaleAge) {
+        return female -> female.getSex() == Person.Sex.WOMAN && female.getAge() >= femaleAge;
```
karma-o, line 130: make non static

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +21,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(string -> Arrays.stream(string.split(DIVIDER)))
+                .mapToInt(Integer::parseInt)
+                .filter(StreamPractice::isEven)
```
karma-o, line 27: same with every other private method
```suggestion
                .filter(this::isEven)
```

karma-o (approved):
Very clean code, you have got a like from me on that! just remove the static and you're good, check the comment on how to do that :)

PR #1043, opened 2024-03-14
jv-stream-solution - V1.0

Nikname2303:


EduarfF (approved):
Well done

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +16,16 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMapToInt(s -> {
+                    String[] parts = s.split(",");
+                    return IntStream.range(0, parts.length)
+                            .map(i -> Integer.parseInt(parts[i].trim()));
```
karma-o, line 23: use `Arrays.stream()` instead, which can be even without braces, like `s -> Arrays.stream(s.split(DELIMITER))` and just map this string to the integer in the next `map` operation, because this is hard to read

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +16,16 @@ public class StreamPractice {
...
+                })
+                .filter(number -> number % 2 == 0)
```
karma-o, line 25: we have some repeating check throughout our class, so we can move this to some method or prediacate and name it  `isEven`

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +34,17 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .mapToDouble(index -> {
+                    int num = numbers.get(index);
+                    if (index % 2 != 0) {
```
karma-o, line 40: this can be moved too, like `isOdd()`

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +75,14 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
...
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> (person.getSex() == Person.Sex.MAN
+                        && person.getAge() >= fromAge
+                        && person.getAge() <= maleToAge)
+                        || (person.getSex() == Person.Sex.WOMAN
+                        && person.getAge() >= fromAge
+                        && person.getAge() <= femaleToAge))
```
karma-o, line 84: lets extract this part to some new predicate inside this class or a method

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +56,11 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getSex() == Person.Sex.MAN
+                        && person.getAge() >= fromAge
+                        && person.getAge() <= toAge)
```
karma-o, line 62: maybe you'll even find a way to reuse some part of this in the previous comment

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,24 @@
...
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= ACCESS_AGE && candidate.isAllowedToVote()
+                && timeInUkraine(candidate.getPeriodsInUkr()) > ACCESS_PERIOD_IN_UKRAINE
+                && candidate.getNationality().equals(NATIONALITY);
+    }
+
+    private int timeInUkraine(String period) {
+        String[] years = period.split("-");
+        int start = Integer.parseInt(years[0]);
+        int end = Integer.parseInt(years[1]);
```
karma-o, line 21: magic values

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +16,16 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMapToInt(s -> {
+                    String[] parts = s.split(",");
```
karma-o, line 21: magic coma

karma-o (changes requested):
Nice, just some minor things

src/main/java/practice/StreamPractice.java
karma-o: not fixed, I mean that `number % 2 == o`

src/main/java/practice/StreamPractice.java
karma-o: same

karma-o (changes requested):
Not all comments are resolved

andrii-hoienko (approved):
Good job!

PR #1046, opened 2024-03-14
Tasks were completed

slizko1:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +21,13 @@ public class StreamPractice {
...
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(n -> n.split(","))
+                .flatMap(Arrays::stream)
+                .map(Integer::valueOf)
+                .filter(n -> n % 2 == 0)
+                .min(Comparator.naturalOrder())
+                .orElseThrow(() -> new NoSuchElementException("Can't get min value from list"));
```
kaiiiseeel, line 30: This condition repeats in your code (40-41 line), so we can move it to the separate method (isEven for example)
`n -> n % 2 == 0`

Also I think, that we can find min without using comparator, btw it will find the minimum number in natural order
` .min(Comparator.naturalOrder()) ---->  .min()`

Plus exception message should look like:
image

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +36,12 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .mapToObj(i -> (i % 2 == 1) ? numbers.get(i) - 1 : numbers.get(i))
+                .filter(num -> num % 2 != 0)
+                .mapToDouble(Integer::doubleValue)
+                .average()
+                .orElseThrow(() -> new NoSuchElementException("No odd numbers found"));
```
kaiiiseeel, line 44: How i already said, it would be great to move our "check if even" logic into separate method.
Also, here you can instantly map to double, without "double-mapping" 
```suggestion
                .mapToDouble(i -> (i % 2 == 1) ? numbers.get(i) - 1 : numbers.get(i))
                .filter(num -> num % 2 != 0)   //replace condition with your own method
                .average()
                .orElseThrow(() -> new NoSuchElementException("No odd numbers found"));
```
Plus i would recommend you to try `getAsDouble()` method instead of `orElseThrow`

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +81,12 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(isWomanValid(femaleAge))
+                .map(Person::getCats)
+                .flatMap(Collection::stream)
```
kaiiiseeel, line 87: Here, I think, you can also combine this 
```suggestion
               .flatMap(person -> person.getCats().stream())
```

kaiiiseeel (changes requested):
Your solution is working for sure, but I think you have to change a few parts of it for better readabillity

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,23 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+
```
andrii-hoienko, line 7:
```suggestion
```

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,23 @@
...
+    private static final int START_YEAR = 0;
+    private static final int END_YEAR = 1;
+
+    @Override
+    public boolean test(Candidate c) {
+        String[] years = c.getPeriodsInUkr().split("-");
+        int yearsInUkraine = Integer.parseInt(years[END_YEAR])
+                - Integer.parseInt(years[START_YEAR]);
+        return c.getAge() >= MIN_AGE
+                && c.getNationality().equals("Ukrainian")
```
andrii-hoienko, line 19: Move string literals to constants

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(n -> Arrays.stream(n.split(",")))
```
andrii-hoienko, line 24: Move string literals to constants

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +99,29 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(new CandidateValidator())
```
andrii-hoienko, line 103: Please make CandidateValidator class level field

andrii-hoienko (changes requested):
Nice, left minor comments

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +21,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(n -> n.split(","))
+                .flatMap(Arrays::stream)
```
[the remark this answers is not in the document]
karma-o: make coma a constant

karma-o (approved):

PR #1047, opened 2024-03-14
feat:initial commit

rel1c-hub:
build & test success

### DIFF
--- a/src/main/java/practice/CandidateValidator.java
+++ b/src/main/java/practice/CandidateValidator.java
@@ -1,5 +1,42 @@
 package practice;
 
-public class CandidateValidator {
+import model.Candidate;
+
+import java.util.function.Predicate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final double MIN_AGE = 35;
+    private static final int MIN_PERIOD_IN_UKR = 10;
+    private static final String NATIONALITY = "Ukrainian";
+    private static final String YEAR_SEPARATOR = "-";
+    private static final int FROM_YEAR = 0;
+    private static final int TO_YEAR = 1;
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= MIN_AGE
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals(NATIONALITY)
+                && getYearsInUkr(candidate.getPeriodsInUkr()) >= MIN_PERIOD_IN_UKR;
+    }
+
+    private static int getYearsInUkr(String candidate) {
+        String[] years = candidate.split(YEAR_SEPARATOR);
+        return Integer.parseInt(years[TO_YEAR]) - Integer.parseInt(years[FROM_YEAR]);
+    }
+
+    @Override
+    public Predicate<Candidate> and(Predicate<? super Candidate> other) {
+        return Predicate.super.and(other);
+    }
+
+    @Override
+    public Predicate<Candidate> negate() {
+        return Predicate.super.negate();
+    }
+
+    @Override
+    public Predicate<Candidate> or(Predicate<? super Candidate> other) {
+        return Predicate.super.or(other);
+    }
     //write your code here
 }
--- a/src/main/java/practice/StreamPractice.java
+++ b/src/main/java/practice/StreamPractice.java
@@ -1,8 +1,12 @@
 package practice;
 
-import java.util.Collections;
+import java.util.Arrays;
 import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import model.Candidate;
+import model.Cat;
 import model.Person;
 
 public class StreamPractice {
@@ -14,7 +18,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMapToInt(e -> Arrays.stream(e.split(","))
+                        .mapToInt(Integer::parseInt))
+                .filter(num -> num % 2 == 0)
+                .min()
+                .orElseThrow(() -> new RuntimeException("Can't get min value from list: "
+                        + numbers));
     }
 
     /**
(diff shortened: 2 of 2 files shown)


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,42 @@
...
+    @Override
+    public Predicate<Candidate> negate() {
+        return Predicate.super.negate();
+    }
+
+    @Override
+    public Predicate<Candidate> or(Predicate<? super Candidate> other) {
+        return Predicate.super.or(other);
+    }
     //write your code here
```
AnyaRatusnyak, line 41: why are you override methods  and, negate, and or ? They are useless in this task

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMapToInt(e -> Arrays.stream(e.split(","))
```
AnyaRatusnyak, line 22: "," - this can be constanta

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,13 @@ public class StreamPractice {
...
+                        .mapToInt(Integer::parseInt))
+                .filter(num -> num % 2 == 0)
```
AnyaRatusnyak, line 24: you can add a new method like  isEvenNumber , and use it here and in next  getOddNumsAverage method

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +33,17 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
```
AnyaRatusnyak, line 36: suggest to make a constant ZEROINDEX = 0;

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +33,17 @@ public int findMinEvenNumber(List<String> numbers) {
...
+                .mapToDouble(index -> {
+                    int num = numbers.get(index);
+                    if (index % 2 != 0) {
+                        num -= 1;
+                    }
+                    return num;
+                })
+                .filter(num -> num % 2 != 0)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
AnyaRatusnyak, line 46: you can make new methods for  subtract 1 from each element on an odd position  and for odd numbers

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +73,16 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
...
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> {
+                    int age = person.getAge();
+                    if (person.getSex().equals(Person.Sex.MAN)) {
+                        return age >= fromAge && age <= maleToAge;
+                    } else {
+                        return age >= fromAge && age <= femaleToAge;
+                    }
+                })
```
AnyaRatusnyak, line 84: I think it will be better for add a new method for filter

AnyaRatusnyak (approved):
Good job!!

src/main/java/practice/CandidateValidator.java
andrii-hoienko: agree, no reason to override this

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,42 @@
...
+
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= MIN_AGE
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals(NATIONALITY)
+                && getYearsInUkr(candidate.getPeriodsInUkr()) >= MIN_PERIOD_IN_UKR;
+    }
+
+    private static int getYearsInUkr(String candidate) {
```
andrii-hoienko, line 22:
```suggestion
    private int getYearsInUkr(String candidate) {
```

src/main/java/practice/StreamPractice.java
andrii-hoienko: agree

src/main/java/practice/StreamPractice.java
andrii-hoienko: agree

src/main/java/practice/StreamPractice.java
andrii-hoienko: I don`t think that is good idea, ZEROINDEX does not add any additional context, so you can left as is

src/main/java/practice/StreamPractice.java
andrii-hoienko: agree, if filter has a lot of content it could be subtracted to method

andrii-hoienko (changes requested):
Please read students comments and resolve them

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +112,11 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        CandidateValidator validator = new CandidateValidator();
```
andrii-hoienko, line 115: Make it class level field

andrii-hoienko (changes requested):

src/main/java/practice/StreamPractice.java
```diff
@@ -10,6 +10,9 @@
 import model.Person;
 
 public class StreamPractice {
+    private static final String SPLITER_FOR_FILE = ",";
+    private CandidateValidator validator = new CandidateValidator();
```
karma-o, line 14: could also be final

karma-o (approved):
GJ! Approved with a comment

PR #1050, opened 2024-03-15
Added nessesary methods

AnyaRatusnyak:


bodya4243 (approved):
All good

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,24 @@
...
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final String SEPARATOR = "-";
+    private static final int STARTINDEX = 0;
+    private static final int ENDINDEX = 1;
+    private static final int MINTERM = 10;
+    private static final int MINAGE = 35;
```
karma-o, line 11: these two "index" names does not really give us more context, so let's think bout some other name to make it more understandable, like what do we get by those indexes? 
Also the constant naming, words are divided by "_"
```suggestion
    private static final int START_INDEX = 0;
    private static final int END_INDEX = 1;
    private static final int MIN_TERM = 10;
    private static final int MIN_AGE = 35;
```

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,24 @@
...
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] periods = candidate.getPeriodsInUkr().split(SEPARATOR);
+        int startOfPeriod = Integer.parseInt(periods[STARTINDEX]);
+        int endOfPeriod = Integer.parseInt(periods[ENDINDEX]);
+        int termOfLivingInUkraine = endOfPeriod - startOfPeriod;
+        return (candidate.getAge() >= MINAGE) && candidate.isAllowedToVote()
+                && candidate.getNationality().equals(NATIONALITY)
+                && termOfLivingInUkraine >= MINTERM;
+    }
```
karma-o, line 23: we can also move the part where we calculate the amount of time that candidate has lived in Ukraine to another method

src/main/java/practice/StreamPractice.java
```diff
@@ -1,31 +1,49 @@
...
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import model.Candidate;
+import model.Cat;
 import model.Person;
 
 public class StreamPractice {
+    private static final String SEPARATOR = ",";
+    private static final int ZEROINDEX = 0;
```
karma-o, line 16: Not necessary, it is usually obvious why we start from 0
```suggestion
```

src/main/java/practice/StreamPractice.java
```diff
@@ -1,31 +1,49 @@
...
      */
+
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(s -> Arrays.stream(s.split(SEPARATOR)))
+                .mapToInt(Integer::parseInt)
+                .filter(this::isEvenNumber)
+                .min()
+                .orElseThrow(() -> new RuntimeException("Can't get min value from list: "
```
karma-o, line 31: let's throw `NoSuchElementException` instead

src/main/java/practice/StreamPractice.java
```diff
@@ -1,31 +1,49 @@
...
      * Given a List of Integer numbers,
      * return the average of all odd numbers from the list or throw NoSuchElementException.
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
+
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(ZEROINDEX,numbers.size())
+                .mapToDouble(index -> oddIndexNumberSubtractOne(numbers, index))
+                .filter(n -> !isEvenNumber((int) n))
```
karma-o, line 43: I do not think we need a cast here

src/main/java/practice/StreamPractice.java
```diff
@@ -34,10 +52,13 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * <p>
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
+
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getAge() >= fromAge && p.getAge() <= toAge)
+                .filter(p -> p.getSex() == Person.Sex.MAN)
```
karma-o, line 59: combine into a single filter operation

src/main/java/practice/StreamPractice.java
```diff
@@ -48,20 +69,27 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
...
      * and each `Cat` having a `name` and `age`),
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
+
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex() == Person.Sex.WOMAN && p.getAge() >= femaleAge)
+                .map(Person::getCats)
+                .flatMap(Collection::stream)
```
karma-o, line 89:
```suggestion
                .flatMap(person -> person.getCats().stream())
```

karma-o (changes requested):
Nice, just a couple more things here, and it should be it

src/main/java/practice/StreamPractice.java
AnyaRatusnyak: Thanks a million)

PR #1055, opened 2024-03-18
done

ChabVlad:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,23 @@
...
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] period = candidate.getPeriodsInUkr().split("-");
+        int startYear = Integer.parseInt(period[0]);
+        int endYear = Integer.parseInt(period[1]);
+        int periodInUA = endYear - startYear;
+        return candidate.getAge() >= 35
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals(ukrNationality)
+                && periodInUA >= 10;
```
olekskov, line 21: magic numbers -> constants with proper naming

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,15 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+
+        return numbers.stream()
+                .flatMap(num -> Arrays.stream(num.split(",")))
+                .map(Integer::parseInt)
+                .flatMapToInt(IntStream::of)
```
olekskov, line 26:
```suggestion
                .mapToInt(Integer::parseInt)
```

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +36,15 @@ public int findMinEvenNumber(List<String> numbers) {
...
-        return 0D;
+        List<Integer> modifiedList = IntStream.range(0, numbers.size())
+                .mapToObj(i -> i % 2 == 0 ? numbers.get(i) : numbers.get(i) - 1)
+                .collect(Collectors.toList());
+
+        return modifiedList.stream()
+                .filter(num -> num % 2 != 0)
+                .mapToDouble(Integer::doubleValue)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
olekskov, line 47: 1) There is no sense to collect stream to open new one again.
2) Too many redundant commands
```suggestion
        List<Integer> modifiedList = IntStream.range(0, numbers.size())
                .map(i -> i % 2 == 0 ? numbers.get(i) : numbers.get(i) - 1)
                .filter(num -> num % 2 != 0)
                .average()
                .orElseThrow(NoSuchElementException::new);
```

olekskov (changes requested):

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,25 @@
...
+    private static final int MIN_PERIOD = 10;
+
+    public CandidateValidator() {
+    }
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] period = candidate.getPeriodsInUkr().split("-");
+        int startYear = Integer.parseInt(period[0]);
+        int endYear = Integer.parseInt(period[1]);
```
okuzan, line 18: not fixed
extract to START_YEAR_INDEX etc

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,25 @@
...
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final String UKR_NATIONALITY = "Ukrainian";
+    private static final int START_AGE = 35;
+    private static final int MIN_PERIOD = 10;
+
+    public CandidateValidator() {
+    }
+
```
okuzan, line 13: no args constructor is created by default if no other is present

src/main/java/practice/StreamPractice.java
okuzan: not fixed, all map/flatMap calls can be combined into one

okuzan (changes requested):

okuzan (approved):

PR #1058, opened 2024-03-23
solve the problem

esidarapp:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,22 @@
...
+                && candidate.getNationality().equals("Ukrainian")
+                && calculateYearsInUkraine(candidate.getPeriodsInUkr()) >= 10;
+    }
+
+    private int calculateYearsInUkraine(String periodsInUkr) {
+        // Предполагается, что periodsInUkr имеет формат "XXXX-YYYY"
+        String[] years = periodsInUkr.split("-");
+        int startYear = Integer.parseInt(years[0]);
+        int endYear = Integer.parseInt(years[1]);
+        return endYear - startYear + 1;
```
okuzan, line 20: why + 1

okuzan, line 20: all literals should be constants

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +67,16 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
...
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> {
+                    if (person.getSex().equals(Person.Sex.MAN)) {
+                        return person.getAge() <= maleToAge && person.getAge() >= fromAge;
+                    } else {
+                        return person.getAge() <= femaleToAge && person.getAge() >= fromAge;
+                        }
+                    }
```
okuzan, line 77: use ternary operator to streamline this

okuzan (changes requested):

okuzan (approved):

PR #1063, opened 2024-03-27
Used stream for methods

YuliiaNisha:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,14 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(l -> Arrays.stream(l.split(",")))
```
kozhukhovsky, line 24: It’s better to give names that aren’t just one letter, even on streams

kozhukhovsky (approved):

PR #1073, opened 2024-04-05
Solved jv-stream task:

Shimeriyu:
Implemented all required methods

### DIFF
--- a/src/main/java/practice/CandidateValidator.java
+++ b/src/main/java/practice/CandidateValidator.java
@@ -1,5 +1,23 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    @Override
+    public boolean test(Candidate candidate) {
+        if (candidate.getAge() < 35) {
+            return false;
+        }
+        if (!candidate.getNationality().equals("Ukrainian")) {
+            return false;
+        }
+        String periods = candidate.getPeriodsInUkr();
+        int from = Integer.parseInt(periods.substring(0,4));
+        int to = Integer.parseInt(periods.substring(5,9));
+        if (to - from < 10) {
+            return false;
+        }
+        return candidate.isAllowedToVote();
+    }
 }
--- a/src/main/java/practice/StreamPractice.java
+++ b/src/main/java/practice/StreamPractice.java
@@ -1,11 +1,19 @@
 package practice;
 
-import java.util.Collections;
+import java.util.Arrays;
+import java.util.Collection;
 import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
 import model.Candidate;
+import model.Cat;
 import model.Person;
 
 public class StreamPractice {
+    private static final Predicate<Integer> isEven = n -> n % 2 == 0;
+
     /**
      * Given list of strings where each element contains 1+ numbers:
      * input = {"5,30,100", "0,22,7", ...}
@@ -14,7 +22,15 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(s -> s.split(","))
+                .flatMap(Arrays::stream)
+                .mapToInt(Integer::valueOf)
+                .filter(isEven::test)
+                .min()
+                .orElseThrow(() -> new RuntimeException("Can't get min value from list: "
+                        + numbers));
+
     }
 
     /**
@@ -23,7 +39,13 @@ public class StreamPractice {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        AtomicInteger indexCounter = new AtomicInteger(0);
+        return numbers.stream()
+                .map(v -> !isEven.test(indexCounter.getAndIncrement()) ? v - 1 : v)
+                .filter(v -> !isEven.test(v))
+                .mapToInt(Integer::intValue)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
     }
 
     /**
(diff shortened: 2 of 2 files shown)


src/main/java/practice/CandidateValidator.java
nowickimj, line 1: Please make all parameters in this class constants

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,23 @@
...
+    @Override
+    public boolean test(Candidate candidate) {
+        if (candidate.getAge() < 35) {
+            return false;
+        }
+        if (!candidate.getNationality().equals("Ukrainian")) {
+            return false;
+        }
+        String periods = candidate.getPeriodsInUkr();
+        int from = Integer.parseInt(periods.substring(0,4));
```
nowickimj, line 16: My recommendation - split by separator rather than hardcoding indexes manually. Code will be more flexible.

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +39,13 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        AtomicInteger indexCounter = new AtomicInteger(0);
```
nowickimj, line 42: Using `AtomicInteger` here is a bit overcomplicated. Use `IntStream.range()`

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +57,11 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex() == Person.Sex.MAN
+                        & p.getAge() >= fromAge
+                        & p.getAge() <= toAge)
+                .collect(Collectors.toList());
```
nowickimj, line 64: `toList()`

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +111,11 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        CandidateValidator candidateValidator = new CandidateValidator();
```
nowickimj, line 114: Could be a constant.

nowickimj (changes requested):

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,31 @@
...
+            return false;
+        }
+        if (!candidate.getNationality().equals(REQUIRED_NATIONALITY)) {
+            return false;
+        }
+        if (to - from < REQUIRED_YEARS) {
+            return false;
+        }
+        return candidate.isAllowedToVote();
+    }
```
okuzan, line 30: just chain all conditions together with &&

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +93,12 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex() == Person.Sex.WOMAN & p.getAge() >= femaleAge)
```
okuzan, line 97: change p to person
avoid short one-letter naming even in stream chains, it's not informative, only makes sense in pure math context

okuzan (changes requested):

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +39,16 @@ public int findMinEvenNumber(List<String> numbers) {
...
+        IntStream oddIndexes = IntStream.range(0, numbers.size())
+                .filter(isOdd::test)
+                .map(i -> numbers.get(i) - 1);
+        IntStream evenIndexes = IntStream.range(0, numbers.size())
+                .filter(isEven::test)
+                .map(numbers::get);
+        return IntStream.concat(oddIndexes, evenIndexes)
+                .filter(isOdd::test)
+                .average()
+                .orElseThrow();
```
nowickimj, line 51:
```java
        .map(index -> index % 2 == 1 ? numbers.get(index) - 1 : numbers.get(index))
        .filter(num -> num % 2 == 1)
        .average()
        .orElseThrow(() -> new NoSuchElementException(
                "There are no matching elements in list: " + numbers
        ));
```

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +79,12 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getAge() >= fromAge
+                        & (person.getSex() == Person.Sex.MAN
+                        ? person.getAge() <= maleToAge : person.getAge() <= femaleToAge)
```
nowickimj, line 85: This predicate is pretty long - please extract it to a variable with a proper name to improve the code readability

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +60,11 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getSex() == Person.Sex.MAN
+                        && person.getAge() >= fromAge
+                        && person.getAge() <= toAge)
```
nowickimj, line 66: This predicate is pretty long - please extract it to a variable with a proper name to improve the code readability

nowickimj (changes requested):

Shimeriyu:
> In my opinion, creating a separate classes for predicates are a bit overkill, but if you want to stick with it, please make sure that you do it properly:
> 
>     1. Make class implement the `Predicate<>` interface
> 
>     2. Create an instance and use it.
> 
> 
> I think that in this case, predicates as constants/variables will make much more sense.

To be honest i did it this way because i don't how i should create predicate as variable when there are non-constans variables for example `int fromAge`

Is something like this acceptable:

`public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
        final Predicate<Person> workablePeoplePredicate = (Person person) ->
                person.getAge() >= fromAge
                & (person.getSex() == Person.Sex.MAN
                ? person.getAge() <= maleToAge : person.getAge() <= femaleToAge);

        return peopleList.stream()
                .filter(workablePeoplePredicate)
                .toList();
    }`
    ?

nowickimj (approved):

PR #1075, opened 2024-04-14
First attempt

jszymanski97:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +21,20 @@ public class StreamPractice {
...
+                .mapToInt(str -> {
+                    try {
+                        return Integer.parseInt(str);
+                    } catch (NumberFormatException e) {
+                        return 0; // or any other default value
+                    }
+                })
+                .filter(n -> n % 2 == 0)
+                .min();
+        return min.orElseThrow(() -> new RuntimeException("Can't get min value from list: "
```
nowickimj, line 36: You used `NoSuchElementException` in other method - why use different exception here?

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +21,20 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        OptionalInt min = numbers.stream()
```
nowickimj, line 24: You don't need to store this value as a variable - make all calculation in single expression:
```java
return numbers.stream()
  ...
  .min()
  .orElseThrow(() -> ...);
```

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +77,12 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> (p.getSex() == Person.Sex.WOMAN && p.getAge()
+                        >= fromAge && p.getAge() <= femaleToAge)
+                        || (p.getSex() == Person.Sex.MAN && p.getAge()
+                        >= fromAge && p.getAge() <= maleToAge))
```
nowickimj, line 84: Please make this predicate a variable with proper name to make code clearer - otherwise it's hard to tell what's happening here. You might as well create two predicates and chain them together - readability is more important than micro-optimization.

nowickimj (changes requested):

src/main/java/practice/StreamPractice.java
jszymanski97: This is what both tasks are asking for :D
for "findMinEvenNumber"    it says "If there is no needed data throw RuntimeException with message..."
and for "getOddNumsAverage" says "return the average of all odd numbers from the list or throw NoSuchElementException."

should I stick with NoSuchElementException for both then ? 
@nowickimj

nowickimj: Good catch - task requirements are sacred, let's stick with them :)

nowickimj (approved):

PR #1080, opened 2024-05-04
Solved described tasks, got some prctice with streams and implemented Predicate logic

vovms:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,23 @@
...
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= MIN_AGE && candidate.isAllowedToVote()
+                && candidate.getNationality().equals(UKRAINIAN)
+                && getDurationInYears(candidate.getPeriodsInUkr()) >= MIN_DURATION;
+    }
+
+    private int getDurationInYears(String periodsInUkr) {
+        String[] data = periodsInUkr.split(PERIOD_DELIMITER);
+        return Integer.parseInt(data[1]) - Integer.parseInt(data[0]);
```
okuzan, line 21: indexes should be constants too

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +88,14 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        Predicate<Person> womenFromAgePredicate = (p) -> p.getSex() == Person.Sex.WOMAN
+                && p.getAge() >= femaleAge;
+        return peopleList.stream()
+                .filter(womenFromAgePredicate)
+                .flatMap(p -> p.getCats().stream())
```
okuzan, line 95: avoid short one letter naming even in stream chains (outside of math context), use person instead of p, for example

okuzan (changes requested):
Good job, minor comments:

sarakhmen (approved):
Good job!

PR #1086, opened 2024-05-16
Implements all methods in StreamPractice.java class. Created validati…

VadymOziPantielieienko:
…on logic in CandidateValidator.java class. Created Predicate classes to avoid long boolean expressions.

### DIFF
--- a/src/main/java/practice/StreamPractice.java
+++ b/src/main/java/practice/StreamPractice.java
@@ -1,8 +1,14 @@
 package practice;
 
-import java.util.Collections;
+import java.util.Arrays;
 import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.OptionalDouble;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import model.Candidate;
+import model.Cat;
 import model.Person;
 
 public class StreamPractice {
@@ -14,7 +20,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(s -> Arrays.stream(s.split(",")))
+                .map(Integer::parseInt)
+                .filter(n -> n % 2 == 0)
+                .min(Integer::compareTo)
+                .orElseThrow(() -> new RuntimeException("Can't get min value from list: "
+                        + numbers));
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +56,12 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        Predicate<Person> sexPredicate = new SexPredicate(Person.Sex.MAN);
+        Predicate<Person> agePredicate = new AgePredicate(fromAge, toAge);
+        return peopleList.stream()
+                .filter(agePredicate)
+                .filter(sexPredicate)
+                .collect(Collectors.toList());
```
Elena-Bruyako, line 64:
```suggestion
                .toList();
```
check all places

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +89,13 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        Predicate<Person> femaleFilterPredictor = new FemaleFilterPredicate(femaleAge);
+        return peopleList.stream()
+                .filter(femaleFilterPredictor)
+                .map(Person::getCats)
```
Elena-Bruyako, line 95:
```suggestion
                .flatMap(p -> p.getCats().stream())
```

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +35,16 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        OptionalDouble average = IntStream.range(0, numbers.size())
+                .mapToObj(i -> i % 2 == 1 ? numbers.get(i) - 1 : numbers.get(i))
+                .filter(n -> n % 2 != 0)
+                .mapToInt(n -> n)
+                .average();
```
Elena-Bruyako, line 42:
```suggestion
                .average()
                .orElseThrow(...;
```

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +35,16 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        OptionalDouble average = IntStream.range(0, numbers.size())
```
Elena-Bruyako, line 38:
```suggestion
        return IntStream.range(0, numbers.size())
```

Elena-Bruyako (changes requested):

src/main/java/practice/StreamPractice.java
VadymOziPantielieienko: here is the question that i did first like u suggesting now but In the checklist is written to do with get/getAsDouble, or I misunderstood something

VadymOziPantielieienko: but i like this more

Elena-Bruyako: Let's leave with orElseThrow

Elena-Bruyako (approved):

PR #1090, opened 2024-05-23
Task completed

apgroupua:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,20 +1,35 @@
...
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
+
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(string -> Arrays.stream(string.split(SPLIT_SYMBOL)))
+                .map(Integer::parseInt)
+                .filter(number -> number % 2 == 0)
+                .min(Comparator.naturalOrder())
```
MAMentorFX, line 30:
```suggestion
                .min()
```

src/main/java/practice/StreamPractice.java
```diff
@@ -1,20 +1,35 @@
...
+                .orElseThrow(()
+                        -> new RuntimeException("Can't get min value from list:" + numbers));
```
MAMentorFX, line 32:
```suggestion
                        -> new RuntimeException("Can't get min value from list: " + numbers));
```

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +88,13 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getSex() == Person.Sex.WOMAN
+                        && person.getAge() >= femaleAge
+                        && !person.getCats().isEmpty())
```
MAMentorFX, line 94:
```suggestion
```
unnecessary work, flatMap will do this work for you

MAMentorFX (changes requested):

src/main/java/practice/StreamPractice.java
apgroupua: .min() Expected 1 arguments (comparator required),
otherwise, it does not compile

Vasul-Pankiv (approved):

PR #1106, opened 2024-06-27
First solution

Wojtas1996PL:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +22,19 @@ public class StreamPractice {
...
+        List<Integer> integerList = correctedList.stream()
+                .map(Integer::parseInt)
+                .toList();
+        Optional<Integer> correctNumber = integerList.stream()
+                .filter(n -> n % 2 == 0)
+                .min(Comparator.comparingInt(a -> a));
+        if (correctNumber.isEmpty()) {
+            throw new RuntimeException("Can't get min value from list: " + correctNumber);
+        }
+        return correctNumber.get();
```
MateuszRostek, line 37: Instead of creating all these redundant variables (List<String>, List<Integer>, Optional<Integer>) you should combine it to a single stream chain. For example like this:

```java
return numbers.stream()
                .map(s -> s.split(","))
                .flatMap(Arrays::stream)
                .mapToInt(Integer::parseInt)
                .filter(this::isNumberEven)
                .min()
                        + numbers));
```

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +43,25 @@ public int findMinEvenNumber(List<String> numbers) {
...
+                .range(0, numbers.size())
+                .filter(i -> i % 2 == 0)
+                .mapToObj(numbers::get)
+                .toList();
+        List<Integer> correctedList = new ArrayList<>(subtractedOddIndexedNumbers);
+        correctedList.addAll(evenIndexedNumbers);
+        OptionalDouble optional = correctedList.stream()
+                .filter(n -> n % 2 != 0)
+                .mapToDouble(n -> Double.parseDouble(String.valueOf(n))).average();
+        return optional.orElseThrow(NoSuchElementException::new);
```
MateuszRostek, line 64: Similar here, you should combine this into the single stream chain. You can also create a private method to check if number is odd/even. Example:

```java
                .map(index -> !isNumberEven(index) ? numbers.get(index) - 1 : numbers.get(index))
                .filter(i -> !isNumberEven(i))
                .average()
```

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +73,11 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex().equals(Person.Sex.MAN))
+                .filter(p -> p.getAge() >= fromAge)
+                .filter(p -> p.getAge() <= toAge)
+                .toList();
```
MateuszRostek, line 80: As AI mentor mentioned:

image

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +92,11 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getAge() >= fromAge)
+                .filter(p -> p.getSex().equals(Person.Sex.MAN) && p.getAge() <= maleToAge
+                        || p.getSex().equals(Person.Sex.WOMAN) && p.getAge() <= femaleToAge)
+                .toList();
```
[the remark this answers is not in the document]
MateuszRostek: .

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +105,15 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
...
-        return Collections.emptyList();
+        List<Person> filteredPersons = peopleList.stream()
+                .filter(p -> p.getSex().equals(Person.Sex.WOMAN))
+                .filter(p -> p.getAge() >= femaleAge)
+                .toList();
+        return filteredPersons.stream()
+                .map(Person::getCats)
+                .flatMap(Collection::stream)
+                .map(Cat::getName)
+                .toList();
```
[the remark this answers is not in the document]
MateuszRostek: AI mentor is right.

MateuszRostek (approved):
Good job! Approved, but see some advice on how you can improve your code:)

PR #1108, opened 2024-06-29
implement methods of StreamPractice class and CandidateValidator class

nikitaabo:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,27 @@
...
+    private static final int END_INDEX = 1;
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] periods = candidate.getPeriodsInUkr().split("-");
+        int start = Integer.parseInt(periods[START_INDEX]);
+        int end = Integer.parseInt(periods[END_INDEX]);
+        return candidate.getAge() >= AGE
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals("Ukrainian")
```
Elena-Bruyako, line 20: create constant for "Ukrainian"

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +51,10 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getAge() >= fromAge && p.getAge() <= toAge
+                        && p.getSex() == Person.Sex.MAN)
+                .collect(Collectors.toList());
```
Elena-Bruyako, line 57:
```suggestion
                .toList();
```
Change for each place

Elena-Bruyako (changes requested):

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,28 @@
 package practice;
 
-public class CandidateValidator {
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
     //write your code here
```
MentorMateAcademy, line 7: clear this comment

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,14 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .map(n -> n.split(","))
+                .flatMap(Arrays::stream)
+                .mapToInt(Integer::parseInt)
```
MentorMateAcademy, line 24: too many methods map

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +82,12 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex() == Person.Sex.WOMAN && p.getAge() >= femaleAge)
+                .map(Person::getCats)
+                .flatMap(Collection::stream)
+                .map(Cat::getName)
```
MentorMateAcademy, line 89: too many methods map

MentorMateAcademy (changes requested):

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return Integer.parseInt(numbers.stream()
+                .map(n -> n.split(","))
+                .flatMap(Arrays::stream)
+                .filter(n -> Integer.parseInt(n) % 2 == 0)
+                .min((o1, o2) -> Integer.parseInt(o1) - Integer.parseInt(o2))
+                .orElseThrow(() -> new RuntimeException("Can't get min value from list "
```
MentorMateAcademy, line 26:
```suggestion
        return Integer.parseInt(numbers.stream()
                .flatMap(s -> Arrays.stream(s.split(",")))
                .mapToInt(Integer::valueOf)
                .filter(n -> n % 2 == 0)
                .min()
                .orElseThrow(() -> new RuntimeException("Can't get min value from list "
```

MentorMateAcademy: try this case

MentorMateAcademy (changes requested):

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +81,12 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex() == Person.Sex.WOMAN && p.getAge() >= femaleAge)
+                .map(p -> p.getCats().stream()
```
okuzan, line 86: even in stream chains, it's better to use longer variable's name than single letter. Naming like n is only appropriate in a math context

okuzan (approved):

PR #1119, opened 2024-07-30
Implement stream processing tasks and candidate validation logic

smagles:


src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +50,10 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getSex() == Person.Sex.MAN)
```
beatarjn, line 54: You can extract both conditions from filter to Predicate and use it in filter.

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +68,11 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getAge() >= fromAge)
```
beatarjn, line 72: The same situation here for Predicate

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +81,13 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+
+        return peopleList.stream()
+                .filter(person -> person.getAge() >= femaleAge
```
beatarjn, line 86: Use Predicate here as well

beatarjn (changes requested):

MateuszRostek (approved):
Good job!

PR #1127, opened 2024-08-09
Task done

nastiapyshak:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,29 @@
...
+        String[] periodsInUkr = candidate.getPeriodsInUkr().split("-");
+        if (periodsInUkr.length != 2) {
+            return false;
+        }
+        return Integer.parseInt(periodsInUkr[1]) - Integer.parseInt(periodsInUkr[0])
+                >= MINIMUM_PERIOD_IN_UKRAINE;
+    }
 }
+
+
```
okuzan, line 29:
```suggestion
```
3 empty lines at the EOF, make it 1

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,29 @@
...
+                && hasLivedInUkraineForTenYears(candidate));
+    }
+
+    private boolean hasLivedInUkraineForTenYears(Candidate candidate) {
+        String[] periodsInUkr = candidate.getPeriodsInUkr().split("-");
+        if (periodsInUkr.length != 2) {
+            return false;
+        }
+        return Integer.parseInt(periodsInUkr[1]) - Integer.parseInt(periodsInUkr[0])
+                >= MINIMUM_PERIOD_IN_UKRAINE;
```
okuzan, line 25: in such cases, don't use boolean literals (true/false), just chain all the conditions together and return boolean expression

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +69,12 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getAge() >= fromAge
+                        && (p.getSex() == Person.Sex.WOMAN
+                                ? p.getAge() <= femaleToAge
```
okuzan, line 75: let's change p to person, use one letter naming only in math context

okuzan (changes requested):
Good job!

Elena-Bruyako (approved):

PR #1135, opened 2024-08-20
The solution of the stream practice.

OlhaHirniak:


src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +102,10 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(c -> new CandidateValidator().test(c))
```
okuzan, line 106:
```suggestion
                .filter(new CandidateValidator())
```

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +81,12 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+
+        return peopleList.stream()
+                .filter(p -> p.getAge() >= femaleAge && p.getSex() == Person.Sex.WOMAN)
```
okuzan, line 86: change p to person, use one letter naming only in math context

okuzan (approved):

PR #1145, opened 2024-08-30
stream api practice solution

Extase67:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +17,14 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(s -> Arrays.stream(s.split(",")))
+                .map(String::trim)
+                .map(Integer::parseInt)
```
MentorMateAcademy, line 23: to many methods map

MentorMateAcademy (changes requested):

Elena-Bruyako (approved):

PR #1149, opened 2024-09-06
added solution

VadikRakitin:


src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +103,11 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        CandidateValidator validator = new CandidateValidator();
+        return candidates.stream()
+                .filter(validator::test)
```
okuzan, line 108:
```suggestion
        return candidates.stream()
                .filter(new CandidateValidator()::test)
```

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +83,11 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getAge() >= femaleAge && p.getSex() == Person.Sex.WOMAN)
+                .flatMap(p -> p.getCats().stream())
+                .map(Cat::getName)
```
okuzan, line 89: use person instead of p for clarity, only use single letter names for math

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,25 @@
...
+        return candidate.getAge() >= MIN_AGE
+                && candidate.isAllowedToVote()
+                && NATIONALITY.equals(candidate.getNationality())
+                && isValidPeriod(candidate.getPeriodsInUkr());
+    }
+
+    private boolean isValidPeriod(String periodsInUkr) {
+        String[] periods = periodsInUkr.split("-");
+        int startYear = Integer.parseInt(periods[0]);
+        int endYear = Integer.parseInt(periods[1]);
```
okuzan, line 22: indexes should be constants too

okuzan (changes requested):

okuzan (approved):

PR #1167, opened 2024-10-21
added solution

Oleksii-Melashchenko:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,71 @@
...
 
-    /**
-     * Given a List of `Person` instances (having `name`, `age`, `sex` and `cats` fields,
-     * and each `Cat` having a `name` and `age`),
-     * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
-     */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getSex().equals(Person.Sex.WOMAN)
```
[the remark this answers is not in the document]
okuzan: == is usually preferrable for enum comparison, - it's compile-time check, and null safe

okuzan (changes requested):

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,70 @@
...
-     * The requirements are: person should be older than 35 years, should be allowed to vote,
-     * have nationality - 'Ukrainian'
-     * and live in Ukraine for 10 years. For the last requirement use field periodsInUkr,
-     * which has following view: "2002-2015"
-     * We want to reuse our validation in future, so let's write our own impl of Predicate
-     * parametrized with Candidate in CandidateValidator.
-     */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        CandidateValidator candidateValidator = new CandidateValidator();
```
ahoienko, line 63: What is the reason to create validator each time this method is called? I suggest you to make it class level variable

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,70 @@
...
-    /**
-     * Given a List of Integer numbers,
-     * return the average of all odd numbers from the list or throw NoSuchElementException.
-     * But before that subtract 1 from each element on an odd position (having the odd index).
-     */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(number -> number % 2 != 0 ? numbers.get(number) - 1 : numbers.get(number))
+                .filter(number -> number % 2 != 0)
```
ahoienko, line 28: Code duplication

ahoienko: (Logic of checking is number is odd could be reused 3 times)

ahoienko (changes requested):
Nice, left minor comments

Elena-Bruyako (approved):

PR #1172, opened 2024-10-25
create streamPractice

jekaGr:


src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +100,10 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        CandidateValidator validator = new CandidateValidator();
+        return candidates.stream().filter(validator)
```
okuzan, line 104:
```suggestion
        return candidates.stream()
                   .filter(new CandidateValidator())
```

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,35 @@
...
+            return false;
+        }
+        if (!candidate.getNationality().equals("Ukrainian")) {
+            return false;
+        }
+
+        if (!candidate.isAllowedToVote()) {
+            return false;
+        }
+        return true;
```
okuzan, line 32: use only one return, change all of this together, extract literals to constants

okuzan (changes requested):

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,37 @@
...
+                return false;
+            }
+        } catch (NumberFormatException e) {
+            throw new RuntimeException("Invalid period format: unable to parse year",e);
+        }
+
+        if (candidate.getAge() < CANDIDATE_OLD_TO) {
+            return false;
+        }
+        if (!candidate.getNationality().equals("Ukrainian")) {
```
okuzan, line 27: extract to NATIONALITY constant

okuzan, line 27: chain all of conditions in `if`s with &&

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,37 @@
...
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int PERIOD_IN_UKRAINE = 10;
+    private static final int CANDIDATE_OLD_TO = 35;
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] periods = candidate.getPeriodsInUkr().split("-");
+
+        try {
+            int period = Integer.parseInt(periods[1]) - Integer.parseInt(periods[0]);
```
okuzan, line 15: indexes to constants too

okuzan (changes requested):

PR #1179, opened 2024-10-29
First version 

MateuszSeler:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,32 @@
...
+        String[] periodInUkrArray = candidate.getPeriodsInUkr().split("-");
+        return candidate.getAge() >= 35
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals("Ukrainian")
+                && (Integer.parseInt(periodInUkrArray[1])
+                - Integer.parseInt(periodInUkrArray[0])) >= 10;
+    }
+
+    @Override
+    public Predicate and(Predicate other) {
```
slade13, line 19: There is no need to override 'and', 'negate' and 'or' methods from Predicate interface without adding any additional logic to them - please remove those overrides.

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,32 @@
...
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] periodInUkrArray = candidate.getPeriodsInUkr().split("-");
```
slade13, line 10: There is no check for the length of 'periodInUkrArray' after splitting the string. An array out-of-bounds exception could occur if the split string does not contain two elements.

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,32 @@
...
+        return candidate.getAge() >= 35
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals("Ukrainian")
+                && (Integer.parseInt(periodInUkrArray[1])
```
slade13, line 14: No exception handling for NumberFormatException if the split string elements are not valid integers.

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,32 @@
...
+                - Integer.parseInt(periodInUkrArray[0])) >= 10;
+    }
+
+    @Override
+    public Predicate and(Predicate other) {
+        return Predicate.super.and(other);
+    }
+
+    @Override
+    public Predicate negate() {
```
slade13, line 24: There is no need to override 'and', 'negate' and 'or' methods from Predicate interface without adding any additional logic to them - please remove those overrides.

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,32 @@
...
+        return Predicate.super.negate();
+    }
+
+    @Override
+    public Predicate or(Predicate other) {
```
slade13, line 29: There is no need to override 'and', 'negate' and 'or' methods from Predicate interface without adding any additional logic to them - please remove those overrides.

slade13 (changes requested):
Please update things highlighted in my comments.

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,21 @@
...
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+
+    @Override
+    public boolean test(Candidate candidate) {
+        if (candidate.getPeriodsInUkr().length() != 9
```
slade13, line 10: The check for the format of 'periodsInUkr' is not robust enough. It only checks the length and the position of the hyphen, but does not check if the years are numbers. It can still pass if the string is '----4----'.

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,21 @@
...
+                || candidate.getPeriodsInUkr().indexOf('-') != 4) {
+            throw new NumberFormatException("something wrong with data input");
```
slade13, line 12: The exception message is not informative enough. It should indicate the exact problem, for instance: 'The periods in Ukraine should follow the format 'yyyy-yyyy''.

slade13 (approved):
Good job! There is only one check missing which is described in the comments but the task is okay!

PR #1187, opened 2024-11-03
First approach to the stream practice task

Pawel-Cp:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,28 @@
...
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= MIN_AGE_OF_THE_CANDIDATE
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals(NATIONALITY)
+                && livedInUkraineFor10Years(candidate.getPeriodsInUkr());
+    }
+
+    private boolean livedInUkraineFor10Years(String periodsInUkr) {
+        String[] years = periodsInUkr.split("-");
```
slade13, line 20: The use of 'split' to parse the 'periodsInUkr' string could lead to unexpected results if the string does not have exactly one hyphen. Consider adding error handling for cases where the string format is not as expected.

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,28 @@
...
+        if (years.length == 2) {
+            int startYear = Integer.parseInt(years[0]);
+            int endYear = Integer.parseInt(years[1]);
```
slade13, line 23: The 'parseInt' method may throw a NumberFormatException if the string cannot be parsed to an integer. Consider adding error handling for this potential issue.

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,28 @@
...
+        return candidate.getAge() >= MIN_AGE_OF_THE_CANDIDATE
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals(NATIONALITY)
+                && livedInUkraineFor10Years(candidate.getPeriodsInUkr());
+    }
+
+    private boolean livedInUkraineFor10Years(String periodsInUkr) {
+        String[] years = periodsInUkr.split("-");
+        if (years.length == 2) {
+            int startYear = Integer.parseInt(years[0]);
```
slade13, line 22: The 'parseInt' method may throw a NumberFormatException if the string cannot be parsed to an integer. Consider adding error handling for this potential issue.

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,28 @@
...
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MIN_AGE_OF_THE_CANDIDATE = 35;
+    private static final String NATIONALITY = "Ukrainian";
+    private static final int MIN_LIVING_IN_COUNTRY = 10;
+
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= MIN_AGE_OF_THE_CANDIDATE
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals(NATIONALITY)
```
slade13, line 15: The use of 'equals' when comparing candidate's nationality to a string constant might lead to a NullPointerException if 'getNationality()' returns null. Consider using 'Objects.equals()' to avoid this potential issue.

slade13 (approved):
Good job, review my comments and think about improving your code if you like.

src/main/java/practice/CandidateValidator.java
Pawel-Cp: done

src/main/java/practice/CandidateValidator.java
Pawel-Cp: done

src/main/java/practice/CandidateValidator.java
Pawel-Cp: done

src/main/java/practice/CandidateValidator.java
Pawel-Cp: done

PR #1205, opened 2024-12-17
solution

ostapzat:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,40 @@
...
+
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= MIN_AGE
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals(REQUIRED_NATIONALITY)
+                && checkTimeLivingInCountry(candidate);
+    }
+
+    private boolean checkTimeLivingInCountry(Candidate candidate) {
```
sarakhmen, line 19: this method is overcomplicated. Try to optimize and refactor it. Also move commas, dashes, indexes etc to constants

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +32,11 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(i -> i % 2 == 1 ? numbers.get(i) - 1 : numbers.get(i))
+                .filter(i -> i % 2 == 1)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
sarakhmen, line 39: let's read common mistakes once again and use more appropriate method here

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +48,10 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getSex().equals(Person.Sex.MAN)
```
sarakhmen, line 52: we can use == with enums

sarakhmen (changes requested):
let's improve your solution ;)

Elena-Bruyako (approved):

PR #1206, opened 2024-12-18
implemented solutions to tasks in StreamPractice class

olieniev:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +20,15 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+
+        return numbers.stream()
+                .map(n -> n.split(","))
```
JJJazl, line 25:
```suggestion
                .map(number -> number.split(","))
```

JJJazl (approved):

PR #1209, opened 2024-12-25
Methods with lambda are implemented

SSXcorp:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,25 @@
...
+
+    @Override
+    public boolean test(Candidate candidate) {
+        List<Integer> years = Arrays.stream(candidate.getPeriodsInUkr().split("-"))
+                .map(Integer::parseInt)
+                .collect(Collectors.toList());
+        return candidate.getAge() >= MINIMAL_AGE
+                && candidate.isAllowedToVote()
+                && candidate.getNationality().equals(NATIONALITY)
+                && (years.get(1) - years.get(0)) >= MINIMAL_YEARS_IN_UA;
```
okuzan, line 23: you can make indexes constants too

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,76 @@
...
-     * We want to reuse our validation in future, so let's write our own impl of Predicate
-     * parametrized with Candidate in CandidateValidator.
-     */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(VALIDATOR)
+                .map(Candidate::getName)
+                .sorted()
+                .collect(Collectors.toList());
```
okuzan, line 74:
```suggestion
                .toList();
```
it's very common, that's why the introduced .toList(), fix everywhere

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,76 @@
...
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getAge() >= fromAge)
+                .filter(person -> {
+                    if (person.getSex() == Person.Sex.MAN) {
+                        return person.getAge() <= maleToAge;
+                    } else {
+                        return person.getAge() <= femaleToAge;
+                    }
```
okuzan, line 55: use ternary operator to streamline this

okuzan (changes requested):
Good job!

src/main/java/practice/StreamPractice.java
SSXcorp: Ok. The answer why I made code like this : On the Mate academy platform task it was impossible to use just toList() construction.

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,27 @@
...
+
+    private static final int MINIMAL_AGE = 35;
+    private static final int MINIMAL_YEARS_IN_UA = 10;
+    private static final String NATIONALITY = "Ukrainian";
+    private static final int START_LIVING_YEAR_INDEX = 0;
+    private static final int END__LIVING_YEAR_INDEX = 1;
+
+    @Override
+    public boolean test(Candidate candidate) {
+        List<Integer> years = Arrays.stream(candidate.getPeriodsInUkr().split("-"))
```
sarakhmen, line 18: can we make a dash a constant as well? fix in all places

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,71 @@
...
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .mapToObj(i -> {
+                    int value = numbers.get(i);
+                    return i % 2 != 0 ? value - 1 : value;
+                })
+                .filter(n -> n % 2 != 0)
+                .mapToDouble(Double::valueOf)
+                .average()
+                .orElseThrow(() -> new NoSuchElementException("No odd numbers found"));
```
sarakhmen, line 35: please read this [common mistakes section](https://github.com/mate-academy/jv-stream-github-practice/blob/main/checklist.md#try-to-use-a-minimum-amount-of-stream-operations) and use more appropriate method

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,71 @@
...
-     * Example: select people of working age
-     * (from 18 y.o. and to 60 y.o. for men and to 55 y.o. for women inclusively).
-     */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getAge() >= fromAge)
+                .filter(person -> person.getSex() == Person.Sex.MAN ? person.getAge() <= maleToAge
+                        : person.getAge() <= femaleToAge)
```
sarakhmen, line 51: in this task we were asked to use the least amount of stream operations. Can we combine these two filters?

sarakhmen (changes requested):
let's slightly improve your solution :)

src/main/java/practice/StreamPractice.java
SSXcorp: What do you mean here? To create my own Predicate for filter or change mapToObject to mapToDouble at the begining like this :

```
                .mapToDouble(i -> {
                    int value = numbers.get(i);
                    return i % 2 != 0 ? value - 1 : value;
                })
                .filter(n -> n % 2 != 0.0)
                .average()
                .orElseThrow(() -> new NoSuchElementException("No odd numbers found"));
}
```

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,70 @@
...
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .mapToDouble(i -> {
+                    int value = numbers.get(i);
+                    return i % 2 != 0 ? value - 1 : value;
+                })
+                .filter(n -> n % 2 != 0.0)
+                .average()
+                .getAsDouble();
```
Sanderfuck, line 34:
```suggestion
                .map(i -> i % 2 == 1 ? numbers.get(i) - 1 : numbers.get(i))
                .filter(n -> n % 2 == 1)
                .average()
                .orElseThrow();
```

Sanderfuck (changes requested):

okuzan (approved):

PR #1238, opened 2025-02-25
First solution

KIR4S3K:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,90 @@
...
+import java.util.NoSuchElementException;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
 import model.Candidate;
+import model.Cat;
 import model.Person;
 
 public class StreamPractice {
-    /**
```
slade13, line 9: Don't remove such comments as they are a part of the task description.

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,90 @@
...
 
-    /**
-     * Given a List of Integer numbers,
-     * return the average of all odd numbers from the list or throw NoSuchElementException.
-     * But before that subtract 1 from each element on an odd position (having the odd index).
-     */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        List<Integer> modifiedNumbers = new ArrayList<>();
+        for (int i = 0; i < numbers.size(); i++) {
```
slade13, line 28: [Don't use any loops](https://github.com/mate-academy/jv-stream-github-practice/blob/main/checklist.md#use-stream-api-to-solve-all-tasks-dont-use-loops) as this is a Stream practice.

slade13 (changes requested):
Everything looks good, but you are not allowed to use any loop in this task. Please rewrite `getOddNumsAverage` method to only use Stream API.

PR #1242, opened 2025-03-05
Tasks ready

KamilSzerszen:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,20 @@
...
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] periodsInUkr = candidate.getPeriodsInUkr().split("-");
+        try {
+            int fromYear = Integer.parseInt(periodsInUkr[0]);
+            int toYear = Integer.parseInt(periodsInUkr[1]);
+            int year = toYear - fromYear;
+            return year > 10;
+        } catch (NumberFormatException e) {
+            throw new RuntimeException("Wrong number format");
```
slade13, line 17: Throwing a generic `RuntimeException` is not ideal. Consider creating a custom exception or using a more specific exception type to provide better context for the error.

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,20 @@
...
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] periodsInUkr = candidate.getPeriodsInUkr().split("-");
```
slade13, line 10: Like the ai mentor said - the method `getPeriodsInUkr()` is assumed to return a string in the format 'fromYear-toYear'. This format should be validated before splitting to ensure it contains exactly one '-' and two valid integers.

slade13 (approved):
Good job, please review my comments.

PR #1256, opened 2025-03-23
firts commit

lisichenkowork:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,19 @@
...
+    private int minAgeOfCandidate = 35;
+    private String nationalityOfCandidate = "Ukrainian";
+    private int minPeriod = 10;
+
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= minAgeOfCandidate
+                && candidate.getNationality().equals(nationalityOfCandidate)
+                && candidate.isAllowedToVote()
+                && Integer.parseInt(candidate.getPeriodsInUkr().split("-")[1])
```
slade13, line 16: The code splits the 'getPeriodsInUkr' string twice, which is inefficient. Consider splitting the string once and storing the result in a variable to improve performance.

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,19 @@
...
+                - Integer.parseInt(candidate.getPeriodsInUkr().split("-")[0]) >= minPeriod;
```
slade13, line 17: The calculation for the period in Ukraine assumes that the years are in chronological order. Consider adding a check to ensure that the start year is less than or equal to the end year to avoid logical errors.

src/main/java/practice/CandidateValidator.java
slade13, line 16: The code assumes that the 'getPeriodsInUkr' method returns a string in the format 'startYear-endYear'. This could lead to a NumberFormatException if the format is incorrect or if the string is not parsable to an integer. Consider adding validation or error handling for this parsing operation.

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +50,10 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(x -> x.getSex().equals(Person.Sex.MAN))
+                .filter(x -> x.getAge() >= fromAge & x.getAge() <= toAge)
```
slade13, line 55: Use `&&` instead of `&` for logical AND operation. `&&` is more efficient as it short-circuits, meaning it stops evaluating as soon as the result is determined.

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +33,12 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .mapToObj(x -> x % 2 == 1 ? numbers.get(x) - 1 : numbers.get(x))
```
slade13, line 37: The logic for subtracting 1 from elements at odd indices is correct, but the method name `getOddNumsAverage` might be misleading. Consider renaming it to something like `getAverageOfOddNumbersAfterAdjustment` for clarity.

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +85,10 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex().equals(Person.Sex.WOMAN) && p.getAge() >= femaleAge)
+                .flatMap(p -> p.getCats().stream().map(c -> c.getName()))
```
slade13, line 90: Consider using method reference `Cat::getName` instead of lambda `c -> c.getName()` for better readability.

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +104,12 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        CandidateValidator candidateValidator = new CandidateValidator();
+
+        return candidates.stream()
+                .filter(candidateValidator::test)
+                .map(c -> c.getName())
```
slade13, line 111: Consider using method reference `Candidate::getName` instead of lambda `c -> c.getName()` for better readability.

slade13 (approved):
Good job, please review my comments to make your code even better!

PR #1278, opened 2025-05-08
Task is solved

AnnaSaienko:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,31 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private int age = 35;
```
liliia-ponomarenko:
```suggestion
    private static final int REQUIRED_AGE = 35;
```

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,31 @@
...
+    private String nationality = "Ukrainian";
+    private int periodsInUkr = 10;
```
liliia-ponomarenko: fix other constant

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,31 @@
...
+        String[] dates = period.split("-");
+        int fromYear = 0;
+        int toYear = 0;
+        try {
+            fromYear = Integer.parseInt(dates[0]);
+            toYear = Integer.parseInt(dates[1]);
+        } catch (NumberFormatException e) {
+            return false;
+        }
+        return (toYear - fromYear + 1) >= periodsInUkr;
```
liliia-ponomarenko: Why do you need +1?

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +33,10 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(n -> n % 2 != 0 ? (numbers.get(n) - 1) : numbers.get(n))
+                .filter(n -> n % 2 != 0)
+                .average().orElseThrow(NoSuchElementException::new);
```
liliia-ponomarenko:
```suggestion
                .average()
                .orElseThrow(NoSuchElementException::new);
```

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +48,11 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> (p.getSex() == Person.Sex.MAN)
+                        && (p.getAge() >= fromAge)
+                        && (p.getAge() <= toAge))
+                .collect(Collectors.toList());
```
liliia-ponomarenko:
```suggestion
                .toList();
```
Fix it everywhere

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +67,11 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> (p.getAge() >= fromAge) && (((p.getSex() == Person.Sex.WOMAN)
+                        && (p.getAge() <= femaleToAge))
+                        || ((p.getSex() == Person.Sex.MAN) && (p.getAge() <= maleToAge))))
```
liliia-ponomarenko: Do you really need all these brackets? For complicated conditional, it is better to use Predicate

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +101,10 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(candidate -> new CandidateValidator().test(candidate))
```
liliia-ponomarenko: better make it a class-level variable. Do you really need a call test method?

liliia-ponomarenko (changes requested):
Good job! Let’s improve your solution ;)

src/main/java/practice/CandidateValidator.java
AnnaSaienko: It was AI-mentor’s comment: “According to typical interpretation, the period should be (end - start + 1) to count both years inclusively.” So I changed my implementation.

Elena-Bruyako (approved):

PR #1288, opened 2025-06-26
first

IgorArsionov:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,22 @@
...
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private final int minYearIn = 10;
+    private final int minYearOld = 35;
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] period = candidate.getPeriodsInUkr().split("-");
```
Elena-Bruyako: all symbols, literals and numbers should be constants

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,16 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(s -> Arrays.stream(s.split(",")))
+                .map(String::trim)
```
```suggestion
```

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +36,15 @@ public int findMinEvenNumber(List<String> numbers) {
...
-        return 0D;
+        List<Integer> modi = IntStream.range(0, numbers.size())
+                .mapToObj(i -> (i % 2 != 0) ? numbers.get(i) - 1 : numbers.get(i))
+                .toList();
+        return modi.stream()
+                .filter(i -> i % 2 != 0)
+                .mapToInt(Integer::intValue)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
+
```
Elena-Bruyako: remove empty line
```suggestion
```

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +56,11 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex() == Person.Sex.MAN)
+                .filter(p -> p.getAge() >= fromAge)
+                .filter(p -> p.getAge() <= toAge)
+                .collect(Collectors.toList());
```
Elena-Bruyako: use just `toList();` 
check all places
```suggestion
                .toList();
```

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +89,13 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(a -> a.getSex() == Person.Sex.WOMAN)
```
Elena-Bruyako: use one filter

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +89,13 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
...
+                .filter(w -> w.getAge() >= femaleAge)
+                .map(Person::getCats)
+                .flatMap(List::stream)
```

Elena-Bruyako (changes requested):

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,24 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private final int minYearIn = 10;
```
Elena-Bruyako: same for other constant
read about constant rules
```suggestion
    private static final int MIN_YEARS_IN = 10;
```

Elena-Bruyako (changes requested):

src/main/java/practice/CandidateValidator.java
IgorArsionov: If I make such a name then it does not perform in tests

Member name 'MIN_YEARS' must match pattern '^[a-z][a-z0-9][a-zA-Z0-9]*$'. [MemberName]

IgorArsionov: I know about the rule for constant names. But tests are not executed if I write according to the rules. Can you explain in more detail, maybe I'm doing something wrong?

IgorArsionov: изображение_2025-06-27_011756854

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +35,14 @@ public int findMinEvenNumber(List<String> numbers) {
...
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        List<Integer> modi = IntStream.range(0, numbers.size())
+                .mapToObj(i -> (i % 2 != 0) ? numbers.get(i) - 1 : numbers.get(i))
+                .toList();
+        return modi.stream()
+                .filter(i -> i % 2 != 0)
+                .mapToInt(Integer::intValue)
+                .average()
+                .orElseThrow(NoSuchElementException::new);
```
liliia-ponomarenko: Make it inline stream

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,24 @@
...
+    private static final int MIN_YEARS = 10;
+    private static final int YEATS_OLD = 35;
+    private static final String SPLITPER = "-";
+    private static final String COUNTRY = "Ukrainian";
+
+    @Override
+    public boolean test(Candidate candidate) {
+        String[] period = candidate.getPeriodsInUkr().split(SPLITPER);
+        int from = Integer.parseInt(period[0].trim());
+        int currentYear = Integer.parseInt(period[1].trim());
```
liliia-ponomarenko:
```suggestion
        int from = Integer.parseInt(period[0]);
        int currentYear = Integer.parseInt(period[1]);
```

liliia-ponomarenko (changes requested):
Let's fix build ;)

Elena-Bruyako (changes requested):
please, fix build

Elena-Bruyako (approved):

PR #1289, opened 2025-07-04
first solution

anpiet:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,27 @@
...
 
-public class CandidateValidator {
-    //write your code here
+import java.util.Objects;
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    @Override
+    public boolean test(Candidate candidate) {
```
slade13, line 9: Silently returning `false `on malformed data may hide bad inputs. You could throw an `IllegalArgumentException `on invalid format so that callers know the data was wrong.

slade13, line 9: Hard‑coding `"Ukrainian"` and `35 `makes it inflexible. Consider exposing them as constructor parameters or constants (e.g. `MIN_AGE`, `NATIONALITY`).

slade13 (approved):
Good job, please review my comments!

PR #1333, opened 2025-10-14
implement StreamPractice class and CandidateValidator predicate class

FranticSkipper:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,75 @@
...
-     * "Can't get min value from list: < Here is our input 'numbers' >"
-     */
+    private static final String SEPARATOR = ",";
+    private static final CandidateValidator candidateValidator = new CandidateValidator();
+
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(n -> Arrays.stream(n.split(SEPARATOR)))
+                .map(el -> Integer.parseInt(el.trim()))
```

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,75 @@
...
-     * Example: select people of working age
-     * (from 18 y.o. and to 60 y.o. for men and to 55 y.o. for women inclusively).
-     */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> {
+                    int maxPersonAge = person.getSex() == Person.Sex.MAN ? maleToAge : femaleToAge;
+                    return (person.getAge() >= fromAge) && (person.getAge() <= maxPersonAge);
```
Elena-Bruyako: why do you need return here?

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,75 @@
...
-     * new Person(«Helen», 42, Sex.WOMAN))`,
-     * select from the List only men whose age is from `fromAge` to `toAge` inclusively.
-     * <p>
-     * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
-     */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> {
+                    return person.getSex() == Person.Sex.MAN
```
Elena-Bruyako: why do you need return here?

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,75 @@
...
-    /**
-     * Given a List of `Person` instances (having `name`, `age`, `sex` and `cats` fields,
-     * and each `Cat` having a `name` and `age`),
-     * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
-     */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> {
+                    return person.getSex() == Person.Sex.WOMAN
```
Elena-Bruyako: remove return

Elena-Bruyako (changes requested):
see comments

src/main/java/practice/StreamPractice.java
FranticSkipper: why don't i need .trim? Suppose the input list can contain elements with spaces, for example [" 1,2, 3   ,4"]. Then .trim would handle this situation, wouldn't it?

src/main/java/practice/StreamPractice.java
FranticSkipper: i used return here to break up the condition. I wanted to avoid having a long expression inside the lambda. I could also use a Predicate, but the task didn’t require it.

FranticSkipper:
7f6f191b9cd73a577e5e86bac1ab4fdf39ec6a5b I removed unnecessary trim and return statements.

KuryloVolodymyr (approved):
good job

PR #1334, opened 2025-10-15
solution

rmaksym1:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +17,16 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers
+                .stream()
+                .map(e -> e.split("\\s*,\\s*"))
```

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +35,17 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(i -> {
+                    int ival = numbers.get(i);
```
Elena-Bruyako: could you simplify solution and remove if statement?

src/main/java/practice/StreamPractice.java
```diff
@@ -50,7 +76,17 @@ public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toA
      */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> isWorkable(p, fromAge, maleToAge, femaleToAge))
+                .toList();
+    }
+
+    private boolean isWorkable(Person p, int fromAge,
```
Elena-Bruyako: private method should be placed after all public methods

Elena-Bruyako (changes requested):
Good job! See comments

KuryloVolodymyr (approved):
good job

PR #1336, opened 2025-10-17
1

AliakseiS97:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,45 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.Objects;
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
```
MateuszRostek, line 7: I think you overcomplicated this class, it could be something like this:

```java
private static final int MINIMUM_AGE = 35;
    private static final int MINIMUM_RESIDENCY_YEARS = 10;

    @Override
        String[] yearsRange = candidate.getPeriodsInUkr().split("-");
        int stayDuration = Integer.parseInt(yearsRange[1]) - Integer.parseInt(yearsRange[0]);

        return candidate.isAllowedToVote() 
                && candidate.getAge() >= MINIMUM_AGE
                && candidate.getNationality().equals(REQUIRED_NATIONALITY)
                && stayDuration >= MINIMUM_RESIDENCY_YEARS;
    }
```

src/main/java/practice/StreamPractice.java
```diff
@@ -34,8 +55,12 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * <p>
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
+
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex() == (Person.Sex.MAN))
+                .filter(p -> p.getAge() >= fromAge && p.getAge() <= toAge)
+                .collect(Collectors.toList());
```
MateuszRostek, line 63:
```suggestion
        return peopleList.stream()
                .filter(p -> p.getAge() >= fromAge
                        && p.getAge() <= toAge
                        && p.getSex() == Person.Sex.MAN)
                .toList();
```

MateuszRostek (approved):
Good job! See minor advice

PR #1337, opened 2025-10-17
Solution

Senjars:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,71 @@
...
-import java.util.Collections;
+import java.util.Arrays;
 import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import model.Candidate;
+import model.Cat;
 import model.Person;
 
 public class StreamPractice {
```
slade13, line 11: Don't remove these block comments they are part of the task.

src/main/java/practice/StreamPractice.java
```diff
@@ -1,80 +1,71 @@
...
-     */
     public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
                                           int maleToAge, List<Person> peopleList) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> (p.getAge() >= fromAge
+                        && p.getAge() <= maleToAge && p.getSex() == Person.Sex.MAN)
+                        || (p.getAge() >= fromAge && p.getAge() <= femaleToAge
+                        && p.getSex() == Person.Sex.WOMAN))
+                .collect(Collectors.toList());
```
slade13, line 50: You can use this instead:

```suggestion
                .toList()
```

slade13 (approved):
Good job!

PR #1340, opened 2025-10-20
Implemented stream tasks with filtering, mapping, and aggregation

tsuki99:


src/main/java/practice/StreamPractice.java
```diff
@@ -1,20 +1,35 @@
...
      * return min integer value. One more thing - we're interested in even numbers.
      * If there is no needed data throw RuntimeException with message
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
+
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(s -> Arrays.stream(s.split(COMMA)))
+                .map(String::trim)
```

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +54,11 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList
+                .stream()
+                .filter(p -> p.getSex() == Person.Sex.MAN
+                        && p.getAge() >= fromAge && p.getAge() <= toAge)
+                .collect(Collectors.toList());
```
Elena-Bruyako: just toList()
same for other places

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +87,12 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex() == Person.Sex.WOMAN && p.getAge() >= femaleAge)
+                .map(Person::getCats)
```

Elena-Bruyako (changes requested):
see comments

src/main/java/practice/StreamPractice.java
tsuki99: why ?
suggestion

src/main/java/practice/StreamPractice.java
tsuki99: why ? in lesson we always used ".collect(Collectors.toList())" to return any List
image

Elena-Bruyako (approved):

PR #1341, opened 2025-10-22
Task solution

KrzysztofPiatkowski:


danon321 (changes requested):
Zadanie poszło do złego mentora. Przekazuje dalej dla mentorów Javy

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +83,12 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex() == Person.Sex.WOMAN && p.getAge() >= femaleAge)
+                .map(p -> p.getCats())
```
mateuszwojtkowiak, line 88: You could use a method reference instead of a lambda on many occasions, and it will be shorter.

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,20 @@
...
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MIN_AGE = 35;
+    private static final int YEARS_IN_UKRAINE = 10;
+    private static final String REQUIRED_NATINALITY = "Ukrainian";
```
mateuszwojtkowiak, line 9: It  should be Nationality instead of Natinality.

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,20 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MIN_AGE = 35;
+    private static final int YEARS_IN_UKRAINE = 10;
```
mateuszwojtkowiak, line 8: It would be better to name it "MIN_YEARS_IN_UKRAINE".

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,20 @@
...
+
+    public boolean test(Candidate candidate) {
+        String[] years = candidate.getPeriodsInUkr().split("-");
+        int start = Integer.parseInt(years[0]);
+        int end = Integer.parseInt(years[1]);
+
+        return candidate.getAge() >= MIN_AGE && candidate.isAllowedToVote()
+            && candidate.getNationality().equals(REQUIRED_NATINALITY)
+            && (end - start) >= YEARS_IN_UKRAINE;
+    }
```
mateuszwojtkowiak, line 19: Overall it will look better if you create a new method in which you will check the needed years in Ukraine instead of doing this in test method. It could look like this:

```public boolean test(Candidate candidate) {
        int yearsInUkraine = calculateYearsInUkraine(candidate.getPeriodsInUkr());

        return candidate.getAge() >= MIN_AGE
                && REQUIRED_NATIONALITY.equals(candidate.getNationality())
                && yearsInUkraine >= MIN_YEARS_IN_UKRAINE;
    }

    private int calculateYearsInUkraine(String period) {
        // your logic here
        return 0;
    }

Wojtas1996PL (approved):
Great job! There some things that can be improved a little, but overall the code looks pretty good!

PR #1346, opened 2025-10-27
added impl of StreamPractice and CandidateValidator

KucherenkoOS:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,43 @@
...
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final int MAX_AGE = 35;
+    private static final int PERIOD_PARTS = 2;
+    private static final int MIN_YEARS_IN_UKRAINE = 10;
+    private static final String NATIONALITY_UKRAINIAN = "Ukrainian";
+
+    @Override
+    public boolean test(Candidate candidate) {
+        if (candidate == null) {
```
Elena-Bruyako, line 14: could you simplify solution and use logical operators?

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +17,13 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(s -> Arrays.stream(s.split(",")))
+                .map(s -> Integer.parseInt(s.trim()))
```

Elena-Bruyako (changes requested):
good job
two comments

Elena-Bruyako (approved):

PR #1354, opened 2025-11-13
stream practice v1

dayren86:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +17,18 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        if (numbers.isEmpty()) {
+            throw new RuntimeException("Can't get min value from list: " + numbers);
+        }
+
+        return numbers.stream()
+                .map(n -> n.split(","))
```
don't use trim

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +17,18 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        if (numbers.isEmpty()) {
+            throw new RuntimeException("Can't get min value from list: " + numbers);
+        }
+
```
```suggestion
```

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +37,11 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        OptionalDouble average = IntStream.range(0, numbers.size())
```
```suggestion
        return IntStream.range(0, numbers.size())
```

src/main/java/practice/StreamPractice.java
```diff
@@ -75,6 +106,12 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * parametrized with Candidate in CandidateValidator.
      */
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        CandidateValidator candidateValidator = new CandidateValidator();
+
```
```suggestion
```

Elena-Bruyako (changes requested):
see comments

Elena-Bruyako (approved):

PR #1355, opened 2025-11-23
practise

szymontwardowski:


src/main/java/model/Cat.java
```diff
@@ -3,6 +3,7 @@
 import java.util.Objects;
 
 public class Cat {
+
```
slade13, line 6: Remove this unwanted/unnecessary change.

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,16 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+
+        return numbers.stream()
+                .flatMap(s -> Arrays.stream((s.split(","))))
+                .map(String::trim)
+                .map(Integer::parseInt)
```
slade13, line 26: You can do it using single map:
`.map(s -> Integer.parseInt(s.trim()))`

src/main/java/practice/StreamPractice.java
```diff
@@ -23,7 +37,19 @@ public int findMinEvenNumber(List<String> numbers) {
      * But before that subtract 1 from each element on an odd position (having the odd index).
      */
     public Double getOddNumsAverage(List<Integer> numbers) {
-        return 0D;
+        return IntStream.range(0, numbers.size())
+                .map(i -> {
+                    ;
```
slade13, line 42: Remove this unnecessary semicolon - as mentioned by AI mentor.

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,38 @@
 package practice;
 
-public class CandidateValidator {
-    //write your code here
+import java.util.Arrays;
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
```
slade13, line 7: Please use constants, as magic numbers are not easy to understand by a developer who sees your code for the first time:
```
    private static final String PERIOD_DELIMITER = ",";
    private static final String YEAR_RANGE_DELIMITER = "-";

    @Override
        boolean meetsBasicRequirements = candidate.getAge() >= MIN_AGE
                && REQUIRED_NATIONALITY.equals(candidate.getNationality());
        ...
    }
}
```

Additionally: `REQUIRED_NATIONALITY.equals(candidate.getNationality())` avoids potential NPE if nationality is `null`.

Now when you have: `candidate.getNationality().equals("Ukrainian");` in case candidate will be null you would get NPE.

slade13 (changes requested):
Good job! Your task is almost perfect - please replace magic numbers with constants.

src/main/java/practice/StreamPractice.java
```diff
@@ -47,7 +47,7 @@ public Double getOddNumsAverage(List<Integer> numbers) {
                     }
                     return value;
                 })
-                .filter(i -> i % 2 != 0)
+                .filter(i -> i % 2 != 0) //
```
slade13, line 50: remove this comment

slade13 (approved):
Good job!

PR #1361, opened 2026-01-05
Stream practice impl

Sidodjji:


src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,17 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        Optional<Integer> min = numbers.stream()
+                .map(s -> s.split(","))
```
Elena-Bruyako: combine flatMap and map

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,17 @@ public class StreamPractice {
...
+                .flatMap(Arrays::stream)
+                .map(Integer::parseInt)
+                .filter(n -> n % 2 == 0)
+                .min(Comparator.naturalOrder());
+        if (min.isPresent()) {
```
Elena-Bruyako: try to avoid if else and use stream methods

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +86,13 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
      * return the names of all cats whose owners are women from `femaleAge` years old inclusively.
      */
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(p -> p.getSex() == Person.Sex.WOMAN
+                        && p.getAge() >= femaleAge)
+                .map(Person::getCats)
```

Elena-Bruyako (changes requested):
see comments

Elena-Bruyako (approved):

PR #1375, opened 2026-02-17
Hw

anurmetov:


anurmetov:
@mateacademy-ai-mentor 
I have reviewed your comments, and i would like to clarify a one point that was misunterstood:

- due to Tests should CandidateValidator.class implement Predicate Interface and not the method in StreamPractice.class, that is why i do not implement Predicate directly in validateCandidates

All other feedback has been adressed.

src/main/java/practice/StreamPractice.java
```diff
@@ -74,7 +119,13 @@ public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
      * We want to reuse our validation in future, so let's write our own impl of Predicate
      * parametrized with Candidate in CandidateValidator.
      */
+
     public List<String> validateCandidates(List<Candidate> candidates) {
-        return Collections.emptyList();
+        return candidates.stream()
+                .filter(new CandidateValidator()) //
```
```suggestion
                .filter(new CandidateValidator())
```

src/main/java/practice/StreamPractice.java
```diff
@@ -59,7 +96,15 @@ public List<Person> getWorkablePeople(int fromAge, int femaleToAge,
...
     public List<String> getCatsNames(List<Person> peopleList, int femaleAge) {
-        return Collections.emptyList();
+        Predicate<Person> isAcceptable = person -> person.getSex() == Person.Sex.WOMAN
+                && person.getAge() >= femaleAge;
+        return peopleList
+                .stream()
+                .filter(isAcceptable)
+                .map(Person::getCats)
+                .flatMap(List::stream)
+                .map(Cat::getName)
```
```suggestion
                .flatMap(p -> p.getCats().stream())
                .map(Cat::getName)
```

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +19,14 @@ public class StreamPractice {
      * "Can't get min value from list: < Here is our input 'numbers' >"
      */
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(str -> Arrays.stream(str.split(",")))
+                .map(String::trim)
+                .map(Integer::parseInt)
+                .filter(integer -> integer % 2 == 0)
```
Elena-Bruyako, line 25:
```suggestion
                .flatMap(input -> Arrays.stream(input.split(",")))
                .map(Integer::parseInt)
                .filter(integer -> integer % 2 == 0)
```

Elena-Bruyako (changes requested):
Good job! 
See some comments

Elena-Bruyako (approved):
Good job

PR #1390, opened 2026-03-24
Candidate validator created and stream practice assingments done

ReduxR:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,35 @@
...
+        }
+        if (!candidate.isAllowedToVote()) {
+            return false;
+        }
+        if (!REQUIRED_NATIONALITY.equals(candidate.getNationality())) {
+            return false;
+        }
+
+        int yearsInUkraine = Stream.of(candidate.getPeriodsInUkr())
+                .map(period -> period.split("-"))
```
Elena-Bruyako: all symbols and numbers should be constants

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,35 @@
...
+
+public class CandidateValidator implements Predicate<Candidate> {
+    private static final String REQUIRED_NATIONALITY = "Ukrainian";
+    private static final int MIN_CANDIDATE_AGE = 35;
+    private static final int MIN_YEARS_IN_UKRAINE = 10;
+    
     //write your code here
+    @Override
+    public boolean test(Candidate candidate) {
+        if (candidate == null) {
```
Elena-Bruyako, line 16: make it simpler to use logical operator instead of if blocks

src/main/java/practice/StreamPractice.java
```diff
@@ -35,7 +48,10 @@ public Double getOddNumsAverage(List<Integer> numbers) {
      * Example: select men who can be recruited to army (from 18 to 27 years old inclusively).
      */
     public List<Person> selectMenByAge(List<Person> peopleList, int fromAge, int toAge) {
-        return Collections.emptyList();
+        return peopleList.stream()
+                .filter(person -> person.getAge() >= fromAge && person.getAge() <= toAge 
+                        && person.getSex().equals(Person.Sex.MAN))
+                .collect(Collectors.toList());
```
Elena-Bruyako: use just `toList() `
fix other places

Elena-Bruyako (changes requested):
Good job
See comments

Elena-Bruyako (approved):

PR #1409, opened 2026-05-13
Practice task

mate-student26:


src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,28 @@
...
+        );
+        return years;
+    }
+
+    @Override
+    public boolean test(Candidate candidate) {
+        return candidate.getAge() >= fromAge
+                && candidate.isAllowedToVote()
+                && nationality.equals(candidate.getNationality())
+                && periodInUkrConvertedToYears(candidate) > MIN_LIVE_PERIOD;
```
slade13: The task says the candidate should have lived in Ukraine for 10 years, and the requirement is usually interpreted as at least 10 years. With the current implementation, a candidate with "2010-2020" would be rejected, because 10 > 10 is false.

Better:

```suggestion
                && periodInUkrConvertedToYears(candidate) >= MIN_LIVE_PERIOD;
```

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,28 @@
...
 
-public class CandidateValidator {
-    //write your code here
+import java.util.function.Predicate;
+import model.Candidate;
+
+public class CandidateValidator implements Predicate<Candidate> {
+    public static final int fromAge = 35;
+    public static final String nationality = "Ukrainian";
+    public static final int MIN_LIVE_PERIOD = 10;
```
slade13: Please use uppercase names for constants and make them private, because they are used only inside CandidateValidator. This improves readability and follows Java naming conventions.

For example:

```suggestion
    private static final int MIN_AGE = 35;
    private static final String REQUIRED_NATIONALITY = "Ukrainian";
    private static final int MIN_LIVE_PERIOD = 10;
```

src/main/java/practice/CandidateValidator.java
```diff
@@ -1,5 +1,28 @@
...
+
+    public int periodInUkrConvertedToYears(Candidate candidate) {
```
slade13: The method periodInUkrConvertedToYears() is used only inside CandidateValidator, so it should be private. Also, the name can be simplified to something like getYearsInUkraine().

src/main/java/practice/StreamPractice.java
```diff
@@ -14,7 +18,14 @@ public class StreamPractice {
...
     public int findMinEvenNumber(List<String> numbers) {
-        return 0;
+        return numbers.stream()
+                .flatMap(n -> Arrays.stream(n.split(",")))
+                .map(String::trim)
+                .map(Integer::parseInt)
+                .filter(n -> n % 2 == 0)
+                .min(Integer::compare)
+                .orElseThrow(() -> new RuntimeException(
+                        "Can't get min value from list: " + numbers));
```
slade13: This works correctly, but you can use primitive streams to avoid unnecessary boxing:
```suggestion
                .flatMap(s -> Arrays.stream(s.split(",")))
                .map(String::trim)
                .mapToInt(Integer::parseInt)
                .filter(n -> n % 2 == 0)
                .min()
                .orElseThrow(() -> new RuntimeException(
                        "Can't get min value from list: " + numbers));
```

slade13 (approved):
Good job, please review my comments and add necessary updates.

